PR #11, opened 2013-01-29
Test and fix for dealing with path regexes with multiple subexpressions

donovanhide:
Hi Rodrigo,

I've been using mux to do some fairly complicated regex matching with multiple named patterns containing multiple subexpressions. This fix and test covers my use case. Let me know if you want any more tests or the same treatment for host regexes.

Cheers,
Donovan.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -194,6 +194,15 @@ func TestPath(t *testing.T) {
 			path:        "/111/222/333",
 			shouldMatch: false,
 		},
+		{
+			title:       "Path route with multiple complex patterns, URL in request does match",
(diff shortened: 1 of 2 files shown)


djgilcrease:
Since you are not actually using the sub expressions (capture groups)

```
/{v1:\\d+(-\\d+)?(:\\d+(-\\d+)?)*}/{v2:\\d+(-\\d+)?(:\\d+(-\\d+)?)*}
```
could be
```
/{v1:\\d+(?:-\\d+)?(?::\\d+(?:-\\d+)?)*}/{v2:\\d+(?:-\\d+)?(?::\\d+(?:-\\d+)?)*}
```
And it will work as expected with the existing code

kisielk:
I believe this will be fixed in #117

PR #22, opened 2013-05-21
Fix %2F inside route parameters.

tcard:
As mux routes match against the request's URL.Path, which get %2F transformed into /, %2F inside parameters don't work as expected. This fixes it by sing the request's RequestURI instead.

[Review on Reviewable](https://reviewable.io/reviews/gorilla/mux/22)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,2 @@
 mux
 ===
-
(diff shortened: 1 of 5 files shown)


kisielk:
7a711bc already contained the changes needed here.

Issue #21, opened 2013-05-17
If a route is matched by everything except the method, router should return a 405 listing the available methods


johnnadratowski:
I have some routes that I only allow certain methods for.  When I try another method for the same URI, I get a 404.  As per the HTTP 1.1 spec, I should get a 405 with the proper Allow header listing the allowed methods.

Reference: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.6

UPDATE:
I also noticed that seemingly unexpected behavior occurs when routes don't match on headers or on queries.  On queries I'd think you'd want to return a 400 with some information about how certain queries aren't allowed.  On headers it might be harder as it seems depending on the header you might want to return a 406, 412, 415, 416, or a 417? I'm not sure with that one.

Feedback is appreciated.  When I get some time, if you guys like the idea, I might make a pull request and do it myself.

kisielk:
Closing in favour of #6

PR #26, opened 2013-06-06
Better error messages

johnnadratowski:
I opened bug #21, which was duplicate of #6 with more information.  This pull request implements _most_ of the suggestions from there.  I didn't get fancy with the header missing requests, so I just ended up using 412 regardless of the headers (some headers, per the protocol, should use different codes, it seems).

I have a good amount of tests in there, and a couple of helper methods.  I have implemented a new NotFound method handler (the default was the one from net/http, which is still used if my code doesn't find an error match).

I saw the response from @nesv about how it would take a lot of refactoring.  I kind of skirted around that by using the context to store error messages/codes that the default NotFound() handler looks for.  If those context variables aren't found, it defaults to the http.NotFound handler.  I thought this would be ok since this package is already tied to gorilla/context since it has the clearing of the context in ServeHTTP.

I hope this is accepted by the project.  I think this is a good way of handling custom errors, and it seems to fit in with being able to write your own matchers, as custom matchers aren't excluded from being able to set context.

If you notice that there is some code that shows up in the diff around clearing the context, it's because I have another pull request for optionally turning off context clearing.  However, I forgot to make a topic branch before I did the implementation for that _facepalm_.  I took that logic out of the code before I started this topic branch, and I double checked that the code from that pull request was taken out, but it still shows up in the diff.

Please let me know if there are any changes you want from me for this pull request.

Thanks!

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -8,19 +8,18 @@ import (
 	"fmt"
 	"net/http"
 	"path"
-    "strings"
+	"strings"
 
 	"github.com/gorilla/context"
 )
 
 var (
-    MatchErrMsgKey string = "matchErrMsgKey"
-    MatchErrCodeKey = "matchErrCodeKey"
+	MuxMatchErrorContextKey string = "MuxMatchErrorContextKey"
 )
 
 // NewRouter returns a new router instance.
 func NewRouter() *Router {
-    return &Router{namedRoutes: make(map[string]*Route)}
+	return &Router{namedRoutes: make(map[string]*Route)}
 }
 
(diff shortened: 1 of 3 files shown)


mttrs:
Is it still alive?

solarfly73:
I'm still getting familiar with the code. Is the differentiation between a missing endpoint vs. a missing method happen the regexp.go area?

solarfly73:
Looks like the problem is r.addMatcher doesn't know the context of the type of thing it's matching on. There's no relationship between MatcherFunc a return code if a match wasn't found, but there needs to be an association.

PR #36, opened 2013-10-07
Middleware (handler wrapping)

scttnlsn:
This adds support for wrapping handler functions with additional functionality.  Example:

```
middleware := func(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
        fmt.Println("before")
        next.ServeHTTP(w, req)
        fmt.Println("after")
    })
}

r := NewRouter().Use(middleware)
```

I find this pattern very useful for implementing auth, logging, etc.  Is this something you'd be interested in merging into `mux`?

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -48,6 +48,8 @@ type Router struct {
 	strictSlash bool
 	// If true, do not clear the request context after handling the request
 	KeepContext bool
+	// Middleware layers to be wrapped around matched handler
+	layers []Middleware
 }
 
 // Match matches registered routes against the request.
@@ -87,6 +89,12 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if !r.KeepContext {
 		defer context.Clear(req)
 	}
+
+	for i := len(r.layers) - 1; i >= 0; i-- {
+		layer := r.layers[i]
+		handler = layer(handler)
+	}
+
 	handler.ServeHTTP(w, req)
 }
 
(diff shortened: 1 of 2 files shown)


mrhwick:
For anyone who is still frustrated by not having a clean way to apply a set of middleware to all Routes in a SubRouter, here's a solution that allows you to specify everything without repeating yourself.

It's a bit complicated, but you can walk the Route tree starting at a certain SubRouter and apply your middleware to all of the Routes below that SubRouter.

```go
func Use(handler http.Handler, middleware ...func(http.Handler) http.Handler) http.Handler {
    for _, m := range middleware {
        handler = m(handler)
    }
    return handler
}

middleware := []func(http.Handler) http.Handler{middleware1, middleware2}

sub := router.PathPrefix("/sub").SubRouter()
sub.Handle("/foo", fooHandler)
sub.Handle("/bar", barHandler)

sub.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error {
	route.Handler(Use(route.GetHandler(), middleware...))
	return nil
})
```

0xdevalias:
In case google or something else leads you here like it did for me.. it seems there is a `Use` function now

https://github.com/gorilla/mux/blob/master/middleware.go#L23-L28 :

```golang
// Use appends a MiddlewareFunc to the chain. Middleware can be used to intercept or otherwise modify requests and/or responses, and are executed in the order that they are applied to the Router.
func (r *Router) Use(mwf ...MiddlewareFunc) {
	for _, fn := range mwf {
		r.middlewares = append(r.middlewares, fn)
	}
}
```

It appears to exist for both routers, and subrouters.

mrhwick:
Thanks for highlighting that new function.

Here's the godoc in case the source tree changes in the future: https://godoc.org/github.com/gorilla/mux#Router.Use

PR #52, opened 2014-04-23
BuildVarsFunc: modify route variables before building the URL from a route

sqs:
This PR introduces a new BuildVarsFunc that can be attached to a route. A BuildVarsFunc allows you to intercept and modify the route variable map passed to `.URL` or `.URLPath` or `.URLHost` before generating the URL.

It is useful when you want to be liberal in matching routes but conservative in constructing them. For example, you might want to normalize all blog post URL slugs to be lowercase, but still match uppercase slugs. Without BuildVarsFunc, you would have to ensure that all callers of `.URL{,Path,Host}` lowercased the slug beforehand. With BuildVarsFunc, you can just call `.URL{,Path,Host}` directly without any additional wrapper functions.

BuildVarsFunc is useful in conjunction with PostMatchFunc at https://github.com/gorilla/mux/pull/53.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -146,6 +146,13 @@ func (r *Router) getRegexpGroup() *routeRegexpGroup {
 	return nil
 }
 
+func (r *Router) buildVars(m map[string]string) map[string]string {
+	if r.parent != nil {
+		m = r.parent.buildVars(m)
+	}
+	return m
+}
+
 // ----------------------------------------------------------------------------
 // Route factories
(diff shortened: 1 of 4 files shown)


beyang:
+1 for making this change!

kisielk:
Okay, after thinking on it some more I decided to merge this. It doesn't complicate things too much and it does indeed seem useful in some cases. Thanks and sorry for taking so long :)

sqs:
Thanks!

PR #54, opened 2014-05-20
Add SkipClean option

whatupdave:
By default paths are run through the cleanPath method which prevents
using fancier paths like /fetch/http://xkcd.com/534

This adds a SkipClean option so that this path isn't redirected to
/fetch/http/xkcd.com/534

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -46,6 +46,8 @@ type Router struct {
 	namedRoutes map[string]*Route
 	// See Router.StrictSlash(). This defines the flag for new routes.
 	strictSlash bool
+	// See Router.SkipClean(). This defines the flag for new routes.
+	skipClean bool
 	// If true, do not clear the request context after handling the request
 	KeepContext bool
 }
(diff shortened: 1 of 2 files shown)


vtolstov:
any news ?

elithrar:
@whatupdave @vtolstov 

This will need to be rebased against master. I'd strongly prefer that this apply to a `*Route` and not a `*Router` so that you can selectively "not clean" specific routes rather than an entire application.

jingweno:
> I'd strongly prefer that this only apply to a *Route and not a *Router so that you can selectively "not clean" specific routes rather than an entire application.

I would want to disable clean path for the entire app since our routes are created dynamically and I don't know ahead of time if the path contains a http URL.

I could rebase this and resolve the conflict.

jingweno:
I rebased it here: https://github.com/gorilla/mux/pull/154.

elithrar:
@jingweno Alright, I'm convinced!

Checklist:

* Rebase against master
* Update comments to indicate that `SkipClean` should only be used when needed, and that failing to clean routes can have unintended consequences
* Tests!

elithrar:
Already added: https://godoc.org/github.com/gorilla/mux#Router.SkipClean

PR #83, opened 2015-02-16
Implement Router.Walk

hjr265:
This pull request introduces the Walk() method we were talking about in Google Groups.. 

The function takes an "iteratee" that satisfies the WalkFunc type. The WalkFunc type is a function that takes a single argument: a Route pointer.

The iteratee is invoked once for every route. However, if it encounters a route that actually holds a sub-router, it recursively calls the Walk() method of the sub-router with the same iteratee. The routes that are handled by a router are always walked in order of insertion (in other words, we just loop over the routes array). Upon encountering a sub-router, it dives in - depth-first. Worth mentioning, the route that actually holds the sub-router is not passed to walkFn. We can still tweak the logic to include it, but it isn't really a "true" route, is it?

The behavior of this method is similar to that of `filepath.Walk`, only simpler. Since `filepath.Walk` attempts to sort the paths of the current directory by name before looping over them. Since not all routes are named in Gorilla, there is nothing really to sort on, and maintaining the order of insertion made more sense.

A test has been added that checks if the order is maintained, and also if the number of routes "walked" on is correct.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -5,6 +5,7 @@
 package mux
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"path"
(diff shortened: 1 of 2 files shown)


hjr265:
We also talked about adding a method that will print the routes in a human-readable format. There is something we might need to discuss before I can implement that. Regarding formatting and the method's signature, that is. I will send an email in the group thread regarding this.

kisielk:
hm ok, how about handling of subrouters? I don't think your current method lets the walk function recognize when it has descended in to a subrouter, does it? For example in the test I don't see the "/z" path appear anywhere.

hjr265:
Ah. Yes. That's true. May be a simple strategy is to not skip those routes that hold the sub-router. That way, the one who is providing the walkFn can account for this.

Also, I was taking a closer look at `filepath.Walk`, and they have a way of preventing the Walk method from descending into sub-directories. May be we should have that in `Router.Walk`. The way they do it is they have a special error that the walkFn can return, and this indicates the Walk function to skip the descend.

kisielk:
Yeah I don't think we should skip the subrouters, also we need some way to signal to the walk function that it's encountered one so it can push to a stack or something like that.

I also agree about the special error to avoid descent.

hjr265:
In that case, shouldn't we also have an indication that it has returned from its descend? I guess that may not be necessary depending on what we push to the stack.

May be we should have a second argument in the walkFn function, which can be a pointer to the current router we are walking on.

kisielk:
yeah that's one possibility. I think a good litmus test is whether or not we can produce a tree with all the paths, including subrouters, in a print function.

hjr265:
So, I made some changes. Two important things to notice here is that we are introducing a SkipRouter value, and changed the WalkFunc function's signature to have three arguments. Here, the `route` argument, is still what it used to be, the current router; the router argument holds a pointer to the current router; and finally, the ancestor argument keeps track of an array of pointers to the routes that lead to this current route/router.

I am not entirely sure if this last argument is an overkill or not, but it certainly is going to be useful, because if someone is using sub-routers and want to use this function, they will always want to keep track of what routes lead to a leaf route in the tree. We are just doing it for them.

This last argument should also allow us build the complete path to any given route from inside a WalkFunc.

djgilcrease:
This does not walk down named routes properly

```go
router.Handle("/", handlers.Index(router))

api_v1 := router.PathPrefix("/api/v1")).Subrouter()

//Define all URLs for Products
products := api_v1.PathPrefix("products")).Name("products").Subrouter()
products_by_id := products.Path("{product_id:[a-f0-9]{8}(?:-[a-f0-9]{4}){3}-[a-f0-9]{12}}").Name("product-by-id").Subrouter()
products_by_name := products.Path("{product_name:[a-z0-9-_~.]+}").Name("product-by-name").Subrouter()
```

The current implementation will only walk / & /api/v1 but none of the /api/v1/products routes

kisielk:
Merged #106 instead.

Issue #62, opened 2014-07-11
URL variables bug for multiple patterns with relatively complex regexp


xdmnl:
Hi,

There's a bug with URL variable with a relatively complex regexp.
For example, let's take the regexp `a|(b/c)` which should match `a` or `b/c`.

If I use the regexp in a path with one pattern, it works. You can test it with these routeTests:

    tests := []routeTest{
      {
        title:       "Path route with single pattern with pipe, match",
        route:       new(Route).Path("/{category:a|(b/c)}"),
        request:     newRequest("GET", "http://localhost/a"),
        vars:        map[string]string{"category": "a"},
        host:        "",
        path:        "/a",
        shouldMatch: true,
      },
      {
        title:       "Path route with single pattern with pipe, match",
        route:       new(Route).Path("/{category:a|(b/c)}"),
        request:     newRequest("GET", "http://localhost/b/c"),
        vars:        map[string]string{"category": "b/c"},
        host:        "",
        path:        "/b/c",
        shouldMatch: true,
      },
    }

But if I try to use it in a path with multiple patterns, it produces weird results.

This test...

    tests := []routeTest{
      {
        title:       "Path route with multiple patterns with pipe, match",
        route:       new(Route).Path("/{category:a|(b/c)}/{product}/{id:[0-9]+}"),
        request:     newRequest("GET", "http://localhost/a/product_name/1"),
        vars:        map[string]string{"category": "a", "product": "product_name", "id": "1"},
        host:        "",
        path:        "/a/product_name/1",
        shouldMatch: true,
      },
    }
...fails because the variables don't match:

    $ go test
    --- FAIL: TestPath (0.00 seconds)
        mux_test.go:810: (Path route with multiple patterns with pipe, match) Vars not equal:
        expected map[category:a product:product_name id:1],
        got      map[category:a product: id:product_name]

And this test...

    tests := []routeTest{
      {
        title:       "Path route with multiple patterns with pipe, match",
        route:       new(Route).Path("/{category:a|(b/c)}/{product}/{id:[0-9]+}"),
        request:     newRequest("GET", "http://localhost/b/c/product_name/1"),
        vars:        map[string]string{"category": "b/c", "product": "product_name", "id": "1"},
        host:        "",
        path:        "/b/c/product_name/1",
        shouldMatch: true,
      },
    }
...also fails because the variables don't match either (but in a different way):

    $ go test
    --- FAIL: TestPath (0.00 seconds)
[...]

PR #89, opened 2015-03-11
Add complementary test for patterns with pipe

gravis:
Sorry, I realized the previous test doesn't make sense if we don't test the other pipe possibility...

refs #62


kisielk:
Thanks. Sorry for the delay :+1:

PR #90, opened 2015-03-13
Basic authorization matcher.

alesstimec:
BasicAuthorization enables one to secure routes using basic username&password-based authorization.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -237,6 +237,12 @@ func (r *Router) BuildVarsFunc(f BuildVarsFunc) *Route {
 	return r.NewRoute().BuildVarsFunc(f)
 }
 
+// BasicAuthorization registers a new route with a basic username&password based authorization matcher.
+// See Route.BasicAuthorization().
+func (r *Router) BasicAuthorization(username, password string) *Route {
+	return r.NewRoute().BasicAuthorization(username, password)
+}
+
 // ----------------------------------------------------------------------------
 // Context
 // ----------------------------------------------------------------------------
--- a/mux_test.go
+++ b/mux_test.go
@@ -5,6 +5,7 @@
 package mux
 
 import (
+	"encoding/base64"
 	"fmt"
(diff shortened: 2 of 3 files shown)


tasdomas:
This looks like a valuable improvement.

kisielk:
Curious as to what the use-case for this is? I can't think of a situation where I'd define a route that only matches one particular username and password. Usually I would wrap the router or a path prefix in another handler that does authentication.

alesstimec:
@kisielk use-case is when you want to secure only a specific endpoint using basic auth, and when you're using subrouters.. example:
root := mux.NewRouter()
secureSubrouter := root.PathPrefix("secure").BasicAuthorization("username", "password").Subrouter()
publicSubrouter := root.PathPrefix("public").SubRouter()

now i can register handlers to each subrouter and have them either secure or public..

avrtno:
Should the server reply with the www authenticate challenge for unauthorized access?  If yes, then this feature does not belong in the router.  If no, what is the use case for this feature?

What are some use cases where the application authenticates against exactly one user name and password where the password is static for the lifetime of the application process? 

Should other authentication methods be supported in the router?  If no, then what's so special about basic auth that it's added?

Use the standard library to parse and format the auth headers:

    func (m *authorizationMatcher) Match(req *http.Request, match *RouteMatch) bool {
        u, p, ok := req.BasicAuth()
        return ok && u == m.username && p = m.pasword
    }

    func newRequestWithCredentials(method, url, username, password string) *http.Request {
	req, err := http.NewRequest(method, url, nil)
	if err != nil {
		panic(err)
	}
        req.SetBasicAuth(username, password)
        return req
    }

kisielk:
I agree with @avrtno, I've never seen a scenario where you'd want to authorize exactly one hardcoded user in a real application. There is usually a lot more complex machinery involved which is beyond the scope of this package.

PR #93, opened 2015-04-22
Fix up doc formatting, use spaces for alignment (tabs are for indentation only).

shurcooL:
This fixes issue where if user has tab width other than 4, some lines will become misaligned. For example, see the package description at https://godoc.org/github.com/gorilla/mux.

image


kisielk:
Thanks :)

Issue #16, opened 2013-04-16
mux headers do not support regex


rbucker:
since accept headers can be complex with all sorts of version information embedded it would be better if it supported regex in the value portion of the pair.

eastwood:
I've added regex support. An example of usage is as follows:
```
	r := mux.NewRouter()
	r.HandleFunc("/", RootHandler).Headers("Content-Type", "application/json.*")
	http.ListenAndServe(":8080", r)
}
```
Now both requests will work:
```
curl -X GET -H "Content-Type: application/json" localhost:8080
curl -X GET -H "Content-Type: application/json;charset=UTF-8" localhost:8080
```
My personal uses were for handling `application/(json|text)`

bluehallu:
@eastwood isn't HeadersRegexp what should be used here?

eastwood:
@bluehallu  That's correct. Above comment is out of data. The documentation should be up to date with this, if anybody requires it

bluehallu:
@eastwood I've tested the code and it behaves as you'd expect, but the docs are in sync with your comment above:

"...Alternatively, you can provide a regular expression and match the header as follows..."

eastwood:
Ah I see :) Good pickup, sorry about the confusion, must of been a legacy commit. I'll clean that up right away to fix the confusion. Alternatively, you can fix the change and make a pull request, the more contributors the merrier!

eastwood:
#125 Should fix this up for us

bluehallu:
I wanted to make sure I wasn't misunderstanding this before doing a PR. All
fixed now so thanks :)

jmassara:
It would be nice to support capturing regex matches in `mux.Vars()` (see my previous comment), similiar to how the `router.HandlerFunc` does it.

kisielk:
@jmassara please open a separate issue for that.

PR #100, opened 2015-07-04
Issue #16: Added regex support for matching headers

eastwood:
Address #16  and added the support for regular expressions.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"regexp"
 
 	"github.com/gorilla/context"
 )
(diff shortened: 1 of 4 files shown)


kisielk:
Hm, I think it's a good idea but potentially problematic for existing users who may be matching literal header values. Interpreting the literal values as regex all of the sudden could mean matching (or not matching) things that weren't intended to be.

kisielk:
So if we want this functionality, I think adding a `HeadersRegexp` function would be better, and then create a new matcher which does that. That will also avoid the overhead of running a regexp instead of a standard string comparison for when it's not needed.

eastwood:
Yea good point. Though, I thought that the `matchMap` function was only used on the header, so I assumed it wouldn't break backwards compatibility.

kisielk:
Well, it *is* only used by the header, but say someone has an existing header matcher that looks for a value like `(foobar)`, `foo*` or `foo+` literally. If we now interpret those patterns as regexps they'll match things they didn't match before, potentially breaking the app.

eastwood:
Yea I understood you're meaning, just didn't have the foresight to see what people would put in their headers (anything). Anyway, I've made alterations, let me know what you think :)

kisielk:
It's a good start, but it's unnecessary to compile the regexp each time the match is attempted, which is what `regexp.MatchString` will be doing. You should `Compile` the regexp just once when the matcher is created and then just use the compiled version's `MatchString` method.

eastwood:
Thanks for the tip, I should have assumed there was an equivalent to `Pattern#compile(String s)` for Golang. Will rebase and try again :)

eastwood:
It's probably not the most idiomatic way to do things. Still learning the best way to do things. Not having inheritance is always fun :P

mux.go
```diff
@@ -327,6 +336,19 @@ func mapFromPairs(pairs ...string) (map[string]string, error) {
...
 }
 
+func mapFromPairsToRegex(pairs ...string) (map[string]*regexp.Regexp, error) {
+	length, err := checkPairs(pairs...)
+	if err != nil {
+		return nil, err
+	}
+	m := make(map[string]*regexp.Regexp, length/2)
+	for i := 0; i < length; i += 2 {
+		regex, _ := regexp.Compile(pairs[i+1])
```
kisielk: I think this should check the error and panic. `regex` will be nil in the error case and the developer will get a panic when the mux tries to match, but it will be hard to trace down the root cause.

If this panics with a message, something like "invalid regular expression: %s" it will be obvious they made a mistake when setting up the routes in the first place.

kisielk: I guess you can just return the error, that will be more in line with the rest of the package's error handling (which badly needs an overhaul, ugh)

eastwood: Yea sounds great. Thanks for the code review, really appreciate :) Will fix this up tomorrow afternoon

eastwood:
@kisielk  Did you decide what you want to do? Do you want me to throw a panic message upon error, or did you want me to let the code to be synonymous with the other style?

kisielk:
Just return the error for now like elsewhere in the code, I'd rather change it all in one fell swoop later.

eastwood:
Yea that's a good idea. I'll be interested in seeing what ideas you have with solving that.

eastwood:
@kisielk  
How'd we go on this one?

kisielk:
Looks good, thanks. As usual GitHub's notification system is a failure as far as telling me things :) Took the @ mention for me to see it...

eastwood:
Yea I figured as much :)
I got a bit delayed with RL, but cheers for code review

PR #101, opened 2015-07-05
Fixing query regexp patterns with quantifiers 

burrbd:
PR adds support for regexp quantifiers for query strings, so that ```Queries("foo", "{v1:[a-z]{3}}")``` will match for ```&foo=bar``` but not for ```&foo=bars```.

The query matcher currently looks for ```foo=[a-z]{3}``` in the entire URL query string (```baz=ding&foo=bars```) and returns a match. PR looks for ```^foo=[a-z]{3}$``` in only ```foo=bars``` and therefore will not match.

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -35,7 +35,6 @@ func newRouteRegexp(tpl string, matchHost, matchPrefix, matchQuery, strictSlash 
 	defaultPattern := "[^/]+"
 	if matchQuery {
 		defaultPattern = "[^?&]+"
-		matchPrefix = true
 	} else if matchHost {
 		defaultPattern = "[^.]+"
 		matchPrefix = false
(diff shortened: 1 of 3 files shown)


regexp.go
```diff
@@ -175,6 +173,18 @@ func (r *routeRegexp) url(values map[string]string) (string, error) {
...
 
+// getUrlQuery returns a single query parameter from a request URL.
+// For a URL with foo=bar&baz=ding, we return only the relevant key
+// value pair for the routeRegexp.
+func (r *routeRegexp) getUrlQuery(req *http.Request) string {
+	keyVal := strings.Split(r.template, "=")
+	if len(keyVal) == 0 {
+		return ""
+	}
+	re := regexp.MustCompile(keyVal[0] + "[^&]*")
```
kisielk: I don't like the idea of compiling a regexp during every match attempt, that's going to hurt performance a lot.

kisielk:
Also what happens if the the query parameter you are matching is not the first one? 

I think this can be achieved by using req.URL.Query() and applying the regexp to each key in the map

burrbd:
Good point and suggestion. Please take a look at my refactored version. Thanks.

kisielk:
ok that seems good. Damn that code (not yours, just the matcher code in general...) is hard to follow though :/

PR #103, opened 2015-07-06
Switch to custom error types

buddhamagnet:
Added some error types at the top of the route file as it makes it easier to see what kind of errors to expect coming back and you can switch on the type rather than having to inspect the string. All tests passing, hope this helps!

### DIFF
--- a/route.go
+++ b/route.go
@@ -12,6 +12,12 @@ import (
 	"strings"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


avrtno:
What is the use case for the application distinguishing between these errors?  

This PR exports error values. It does not create custom error types.

kisielk:
Not going to merge this since I want to change up the error handling strategy in mux.

PR #104, opened 2015-07-08
Add URLPathTemplate to Route …

djgilcrease:
to make it easier to generate a API page that lists all routes by template

This works well with https://github.com/gorilla/mux/pull/83 so you can do something like

```go
func Index(router *mux.Router) *index {
    return &index{router}
}

type index struct {
    router *mux.Router
}

func (h *index) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
    h.router.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error {
        path, _ := route.URLPathTemplate()
        fmt.Fprintf(rw, "%s\n", path)
        return nil
    })
}```

### CODE REVIEW

**Comment by elithrar** [comment]
I'm going to close this one out as I don't think it's a useful addition to the public API, but I'm happy to be convinced otherwise if others have a need for it.

**Comment by djgilcrease** [comment]
Obviously I find it useful. I use it in every rest service I have written with gorilla to serve up a simple api doc. I also use it heavily for instrumentation with new relic to ensure it is reporting consistently.

This also allows @tleyden to do what he is requesting in issue #129

File: route.go [hunk 1]
```diff
@@ -508,6 +508,19 @@ func (r *Route) URLPath(pairs ...string) (*url.URL, error) {
 	}, nil
 }
 
+// URLPathTemplate returns the template used to match against for the route
```
elithrar: Can you expand on this? What is this useful for?

elithrar:
@djgilcrease 

If you can address my comment and add tests I'll re-consider. Really big on having tests for this since mux sees a lot of PRs.

djgilcrease:
Updated comments, and added a new testTemplate method that is called everywhere testRoute is called.

route.go
```diff
@@ -532,6 +532,30 @@ func (r *Route) URLPath(pairs ...string) (*url.URL, error) {
 	}, nil
 }
 
+// GetPathTemplate and GetHostTemplate returns the template used to match against for the route
```
elithrar: `// GetPathTemplate and GetHostTemplate return the template used to build the route match.`

route.go
```diff
@@ -532,6 +532,30 @@ func (r *Route) URLPath(pairs ...string) (*url.URL, error) {
...
+// This is userful for building simple REST API documentation,
+// and instrumentation for services like New Relic to ensure consistent reporting
+// The route must have a path defined.
```
elithrar: // This is useful for building simple REST API documentation and for instrumentation
// against third-party services.
// An error will be returned if the route does not define a path.

route.go
```diff
@@ -532,6 +532,30 @@ func (r *Route) URLPath(pairs ...string) (*url.URL, error) {
...
+	if r.err != nil {
+		return "", r.err
+	}
+	if r.regexp == nil || r.regexp.path == nil {
+		return "", errors.New("mux: route doesn't have a path")
+	}
+	return r.regexp.path.template, nil
+}
+
+func (r *Route) GetHostTemplate() (string, error) {
```
elithrar: Need its own godoc string.

```go
```go
// GetHotsTemplate returns the template used to build the
// route match.
// An error will be returned if the route does not define a host.
```

djgilcrease:
All updated based on latest feedback

elithrar:
Merged. Thanks!

PR #106, opened 2015-07-09
Update the walk method  …

djgilcrease:
…to walk matchers so it walks the full list of routers and child routers

This is just an update of the work done in https://github.com/gorilla/mux/pull/83 to properly account for subrouters


kisielk:
This looks good, thanks.

PR #108, opened 2015-07-16
fix for empty query

burrbd:
This PR is a solution to the issue raised by @epelc in https://github.com/gorilla/mux/commit/0c9d5c08c1d6103f2ced11750a195fc94902d3fc#commitcomment-12196652

Not sure if it's the right approach but at least something for discussion. There should be a more idiomatic/explicit way to test if the ```defaultPattern``` should be added to the ```pattern``` but this was an easy way to do it.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -588,6 +588,15 @@ func TestQueries(t *testing.T) {
 			path:        "",
 			shouldMatch: false,
 		},
+		{
+			title:       "Queries route with empty value, should match",
+			route:       new(Route).Queries("foo", ""),
+			request:     newRequest("GET", "http://localhost?foo=bar"),
+			vars:        map[string]string{},
+			host:        "",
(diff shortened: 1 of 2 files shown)


kisielk:
Looks good, thanks.

Really need to refactor addRegexpMatcher in to three separate functions at some point, it's getting kind of ridiculous with all the flags....

burrbd:
@kisielk Have another branch to refactor the addRegexpMatcher, which I'll submit soon. Currently experimenting with lexical analysis and parsing the template (which may be overkill) as opposed to looping over the brace positions and splitting strings (which IMO is difficult to follow).

kisielk:
@burrbd sounds good. I figured the whole thing could be a lot simpler if there was a separate function to create a matcher for each, since really there is not much overlap between the logic there...

Issue #109, opened 2015-07-17
Routing based upon queries matches incorrect query.


dkumor:
```go
subrouter.HandleFunc("/blah/blah", myfunc1).Methods("GET").Queries("i1", "{i1}")
subrouter.HandleFunc("/blah/blah", myfunc2).Methods("GET").Queries("t1", "{t1}")
```

- Expectation: `GET /blah/blah?t1=0` will call `myfunc2`
- What happens: `myfunc1` is called

The issue appeared within the last 24 hours - the code functioned according to the expectations yesterday (7/15).

kisielk:
@burrbd any chance you can have a look at this?

burrbd:
I think the query parameter ```defaultPattern``` change from #108 is causing this.

Queries("t1", "{t1}") creates ```^t1=[^?&]*$```. The string from the request to match against will be ```t1=``` even if the parameter wasn't present in the URL.

dkumor:
The merged PR fixed the issue. Thanks a lot!

PR #111, opened 2015-07-17
Issue #109

burrbd:
Fix for https://github.com/gorilla/mux/issues/109.

Unfortunately we broke the query matcher when the query's ```defaultPattern``` was changed in https://github.com/gorilla/mux/pull/108.

In this PR the ```getUrlQuery``` method loops over query parameters and only returns a string to match if the the parameter exists in the request (which is how it should have been in the first place).

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -598,6 +598,24 @@ func TestQueries(t *testing.T) {
 			shouldMatch: true,
 		},
 		{
+			title:       "Queries route with empty value and no parameter in request, should not match",
+			route:       new(Route).Queries("foo", ""),
+			request:     newRequest("GET", "http://localhost"),
+			vars:        map[string]string{},
+			host:        "",
+			path:        "",
+			shouldMatch: false,
+		},
+		{
+			title:       "Queries route with empty value and empty parameter in request, should match",
+			route:       new(Route).Queries("foo", ""),
+			request:     newRequest("GET", "http://localhost?foo="),
+			vars:        map[string]string{},
+			host:        "",
+			path:        "",
+			shouldMatch: true,
+		},
+		{
 			title:       "Queries route with overlapping value, should not match",
 			route:       new(Route).Queries("foo", "bar"),
 			request:     newRequest("GET", "http://localhost?foo=barfoo"),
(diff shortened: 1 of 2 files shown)


mux_test.go
```diff
@@ -606,6 +624,24 @@ func TestQueries(t *testing.T) {
 			path:        "",
 			shouldMatch: false,
 		},
+		{
+			title:       "Queries route with no parameter in request, should not match",
```
burrbd: This is the test that should cover #109.

regexp.go
```diff
@@ -186,9 +186,13 @@ func (r *routeRegexp) getUrlQuery(req *http.Request) string {
 	if !r.matchQuery {
 		return ""
 	}
-	key := strings.Split(r.template, "=")[0]
-	val := req.URL.Query().Get(key)
-	return key + "=" + val
+	templateKey := strings.Split(r.template, "=")[0]
```
kisielk: use `SplitN` here with a max of 2

kisielk: I just did this quickly :)

kisielk:
Thanks for the fix.

It really seems unnecessary to put the key and = in to the template now, I have to get around to refactoring this code and creating a dedicated query matcher which just has a key and value field to avoid all of this mess.

burrbd:
Was a bad thing to break. Hopefully those extra tests will help with any refactoring. It would be nice if the refactoring could handle ```foo=1&foo=2&foo=3```. Might need a new queries variable along with ```mux.Vars(request)``` to do this.

kisielk:
You could already extract that from `r.URL.Queries()["foo"]` inside your handler if that's what you need...

burrbd:
True. But it's not great that in this case ```vars["foo"] ``` would be equal to a string (the first value) as opposed to a slice.

And the matcher should test the value of each ```&foo``` rather than just one item ... I think? O_o

kisielk:
Yeah, it's unfortunate about vars but we can't change the API.

But yes, I the matcher should check each foo and just return the first match, which is what it's doing now.

kisielk:
Actually, that's not quite right.. it's only attempting the match on the first query... *sigh*. It seems like we'd probably have to do a fair bit of refactoring to get the desired behaviour with multiple query items with the same name.

PR #112, opened 2015-07-22
Export setVars

zimmski:
I need setVars for testing purposes exported. I am testing single handlers and would like to use the Vars function inside them.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -84,7 +84,7 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	var handler http.Handler
 	if r.Match(req, &match) {
 		handler = match.Handler
-		setVars(req, match.Vars)
+		SetVars(req, match.Vars)
 		setCurrentRoute(req, match.Route)
 	}
 	if handler == nil {
@@ -318,7 +318,8 @@ func CurrentRoute(r *http.Request) *Route {
 	return nil
 }
 
-func setVars(r *http.Request, val interface{}) {
+// SetVars sets the route variables for the current request.
+func SetVars(r *http.Request, val interface{}) {
 	context.Set(r, varsKey, val)
 }
 


kisielk:
I think there's other ways to test this that would be just as effective or more, without needing to expand the API of the package.

zimmski:
It seems to me that it is the easiest. Create a http.Request, add the vars to mux.SetVars and then forward it into the handler. There is actually no other mux code involved.
I am up for suggestions.
(thanks for the awesome package btw, using it for what seems ages)

kisielk:
The way I usually test is create a mock server using the `net/http/httptest` package and actual create a router instance. Then I send requests to the server which will call the appropriate handler and produce the vars I want to test. That way the routing can be tested as well.

zimmski:
Yes, I always did that too, but I noticed that testing controller handlers could be much simpler and resource friendly by simply calling them directly with a self-made http.Request instance. That way there are no unneeded server and router instances. Also then there is no need for knowing the URI. Do not get me wrong, I still test the routers. However, the dependencies of the controllers and handlers in the project are completely mockable and therefore can run perfectly fine in parallel tests. Instantiating a server and a route for every handler is just inefficient and that is the reason why I made the change in the first place.

Arkan:
+1 for this change too!

kisielk:
@elithrar any thoughts on this?

It does add to the API a bit, but then again it doesn't add more methods to Route or Router...

My only qualm is that people may abuse it for things other than testing and that may complicate some uses of the package later...

elithrar:
I'd prefer to keep it private—since writing your own test helper should be a fairly simple case of calling `context.Set`.

kisielk:
Ah right, I didn't even think of that :) Right then, going to close this.

zimmski:
I thought so too but that does not work. The call would mimic `context.Set(r, varsKey, val)` and `varsKey` is of type `contextKey`. Both are not exported, so I thought I could just use the value, which is always "0". However, the key is of type interface{}, so the type of the key value becomes important. To get `context.Get` to work, you need a `mux.contextKey(0)`.

cam-stitt:
Aware that this has been closed and is an older issue, but I think @zimmski makes a pretty good point that you can't use context.Set because of the private `contextKey` type as well. If I am testing the handlers using `httptest.NewRecorder` it would be great to have one of these exported.

bhirbec:
I gave some thoughts about this too and I join @zimmski and @cam-stitt. My handlers are wrapped into several middlewares and using the mux/server approach would require me to write more code. As an example, my authentication middleware requires some specific HTTP headers which I could get rid of by calling the handler directly. 

It's really unfortunate that there's no way to directly call the handler. I will be glad to provide a PR to open SetVar().

PR #113, opened 2015-07-24
Refactor and create custom errors

buddhamagnet:
1. Create custom errors.
2. Cache length for loop counter.
3. Capture blank string in variable.

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -183,8 +183,9 @@ func (r *routeRegexp) url(values map[string]string) (string, error) {
 // For a URL with foo=bar&baz=ding, we return only the relevant key
 // value pair for the routeRegexp.
 func (r *routeRegexp) getUrlQuery(req *http.Request) string {
+	blank := ""
 	if !r.matchQuery {
-		return ""
+		return blank
 	}
 	templateKey := strings.SplitN(r.template, "=", 2)[0]
 	for key, vals := range req.URL.Query() {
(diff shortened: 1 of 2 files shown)


regexp.go
```diff
@@ -183,8 +183,9 @@ func (r *routeRegexp) url(values map[string]string) (string, error) {
 // For a URL with foo=bar&baz=ding, we return only the relevant key
 // value pair for the routeRegexp.
 func (r *routeRegexp) getUrlQuery(req *http.Request) string {
+	blank := ""
```
avrtno: This change hurts the readability of the code.  The variable `blank` is used as a constant. Even if the declaration is changed to `const blank = ""`, the change hurts readability. It's one more line of code to understand with no extra value.

route.go
```diff
@@ -12,6 +12,10 @@ import (
 	"strings"
 )
 
+var ErrorNoHostOrPath = errors.New("mux: route doesn't have a host or path")
+var ErrorNoHost = errors.New("mux: route doesn't have a host")
+var ErrorNoPath = errors.New("mux: route doesn't have a path")
```
avrtno: What is the use case for allowing the application to distinguish between these errors?  If there's no use case, then it's best to leave them out the exported interface for the package. Less is more.

If these variables are added to the package, then there should be documentation describing where these errors are returned.  Without documentation, these errors are noise.

PR #116, opened 2015-08-05
Add note about the availability of CurrentRoute.

brocaar:
For a middleware I'm working on I wanted to use ``CurrentRoute`` and was surprised it always returned ``nil``. Then I realised it is only set just before the handler will be called and will be cleared directly after (on purpose) here: https://github.com/gorilla/mux/blob/master/mux.go#L98

I think it would be a good idea to add a small note, to take the expectation away that ``CurrentRoute`` can be used outside the handler.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -312,9 +312,10 @@ func Vars(r *http.Request) map[string]string {
 }
 
 // CurrentRoute returns the matched route for the current request, if any.
-// Note: this only works when called inside the handler of the matched route
-// because it uses context.Get() which will be cleared after executing the
-// handler.
+// This only works when called inside the handler of the matched route
+// because the matched route is stored in the request context which is cleared
+// after the handler returns, unless the KeepContext option is set on the
+// Router.
 func CurrentRoute(r *http.Request) *Route {
 	if rv := context.Get(r, routeKey); rv != nil {
 		return rv.(*Route)
--- a/mux.go
+++ b/mux.go
@@ -312,6 +312,9 @@ func Vars(r *http.Request) map[string]string {
 }
 
 // CurrentRoute returns the matched route for the current request, if any.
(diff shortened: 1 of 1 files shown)


kisielk:
So you were using middleware which wrapped the Router? It should be available in middleware which wraps a handler since mux will call it...

brocaar:
I'm using ``mux`` with ``negroni``. If I understand it correctly, when you use ``negroni.New(Middleware, ...)`` and then ``negroni.UseHandler(r)`` to set the router, the middlewares will be called before the mux handler. So before calling the next middleware, ``mux`` hasn't set ``context`` yet, after the call the ``context`` is already cleared:

```
Middleware 1
    -> Middleware 2
        -> Gorilla mux
            * Set current handler + vars (context)
              -> Call handler
            * Context.Clear
```

kisielk:
Yes, that's correct.

If you want to do it that way, you can enable KeepContext and then manually clear with another middleware that wraps the whole stack.

mux.go
```diff
@@ -312,6 +312,9 @@ func Vars(r *http.Request) map[string]string {
 }
 
 // CurrentRoute returns the matched route for the current request, if any.
+// Note: this only works when called inside the handler of the matched route
```
kisielk: Change the wording a bit. Remove "Note:" and instead of "it uses context.Get()" say:

"the matched route is stored in the request context which is cleared after the handler returns, unless the `KeepContext` option is set on the Router.`

brocaar:
Yep, I saw that in the code. Just created this pull-request since I think it would be good to add a small comment to the documentation about this behaviour :-) Potentially, more people run into this issue.

mux.go
brocaar: Done

PR #119, opened 2015-08-10
Changed the readme to embed "www.domain.com" into inline code blocks

ineentho:
The "www.domain.com" is currently automatically linked to domain.com, which is undesired.


kisielk:
The examples should really use example.com instead. Do you mind just replacing all occurrences of domain.com or other URLs with that?

PR #120, opened 2015-08-11
Reduce variable name restriction

shkw:
I was using hyphenated names for path variables and they had been working well before the commit 8965bfef36e398a40fa4afbe4e22eea61cb8db86.
But after the commit was pushed, hyphenated variables names cause a regexp syntax error and it is causing a problem to me.

I believe that mux should keep the support of such variable names, hence I made some changes to achieve it.
I'd appreciate it if my changes could be accepted.

If the naming convention was intentionally revised, my changes can be ignored but it is really helpful to document the naming convention somewhere in README.md instead.


kisielk:
Seems reasonable. Breaking hyphenated names was certainly un-intentional, so thanks for finding that.

Any particular reason you chose to remove the `name != ""` checks in `setMatch` though?

shkw:
Thanks for the prompt response.

> Any particular reason you chose to remove the name != "" checks in setMatch though?

The `name == varGroupName(varName)` checks ensure the `name` is not empty since `varGroupName` func always returns a non-empty string, so I think the `name != ""` checks can be removed.

kisielk:
That's true, but you can skip the `Sprintf` call entirely if the name is empty since you know it will never match. It would probably afford a slight difference in performance when there are nested subexpressions (and hence blank names).

For that matter, I'd rather `varGroupName` didn't use `fmt.Sprintf` since that requires additional interface to concrete type conversions. I think a string concatenation with the result of `strconv.FormatInt` would be preferable.

shkw:
I totally agree with you.
I will make changes accordingly.

shkw:
I made changes as you suggested.
Could you please review the code again?

kisielk:
That looks good, thank you. Also thanks a lot for the thorough test cases, it's much appreciated.

Issue #126, opened 2015-09-08
Regex captures from HeadersRegexp


jmassara:
It would be very helpful to support capturing regex matches in `mux.Vars()`, similiar to how the `router.HandlerFunc` does it.

Example:

```go
func main() {
	r := mux.NewRouter()
	r.HandleFunc("/myapp", MyHandler).
		Methods("GET").
		HeadersRegexp("Accept", "application/vnd.myapp-v{version:[0-9]+}+json")
	http.Handle("/", r)
	log.Fatalln(http.ListenAndServe(":8080", r))
}

func MyHandler(w http.ResponseWriter, r *http.Request) {
	vars := mux.Vars(r)
	log.Printf("%v", vars)
    // would display the version variable
}
```

pH14:
Working on this--I have something that kinda works but I did some horrible things to the existing code to make it happen :smile:

I'll work on cleaning it up and then send in a PR

kisielk:
Awesome. Thanks :+1:

elithrar:
Closing due to inactivity.

PR #127, opened 2015-09-24
HeadersRegexp stores regexp captures in mux's Vars

pH14:
Rejigged `HeadersRegexp` a bit so that it can store regexp captures into mux's vars, per the suggestion from https://github.com/gorilla/mux/issues/126

In terms of it working just right, the bits I'm least certain about are my changes to `newRouteRegexp`. There is also a fair bit of if-else logic in various places that cycle between the different matchers cases, so those clauses could use some extra scrutiny where I had to insert changes, like in `addRegexpMatcher`.

One design consideration: the other matchers like `Path`, `Queries`, etc. require regexp strings to have braces, so `addRegexpMatcher` and related functions expect this. Since `HeadersRegexp` accepted regexp strings without braces before, I set it up so that it inserts regular matchers for non-braced strings, and uses a regexp matcher for braced strings. I don't know if using both seems too hacky or not, but I didn't want to break existing usage of the function.

What do you think @kisielk?

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -537,6 +537,24 @@ func TestHeaders(t *testing.T) {
 			path:        "",
 			shouldMatch: true,
 		},
+		{
+			title:       "Headers route, regex header values to match with captured vars",
+			route:       new(Route).HeadersRegexp("foo", "{v1:ba[zr]}"),
+			request:     newRequestHeaders("GET", "http://localhost", map[string]string{"foo": "baz"}),
+			vars:        map[string]string{"v1": "baz"},
+			host:        "",
+			path:        "",
+			shouldMatch: true,
+		},
+		{
+			title:       "Headers route, regex header values to match with several captured vars",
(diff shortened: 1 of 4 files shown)


kisielk:
I am a bit concerned about the `addRegexpMatcher` modification. That function just simply does too much and is really complicated to maintain, I'd rather not grow it. I hope to be able to do some refactoring and split it up, to have a separate function for each kind of matcher. I have a local branch that does some of this, but it's not fully functional yet...

pH14:
That seems reasonable, I can wait and merge my changes into what you come up with

kisielk:
I pushed the branch up to GitHub, it's in the matcher-refactor branch. It looks like so far I got Queries to create its own matcher without relying on that big mess of a function. The other types of matches still need to happen.

Not sure if you want to take a look at it or not, but I'd appreciate some help working on the refactor if you're interested. I don't have a whole lot of time to devote to it at the moment.

pH14:
Reworking that might be biting off more than I can chew at the moment, but once I have a bit more time I'd be happy to take a look

PR #133, opened 2015-11-08
Correct a printf verb type

mttrs:
% go vet
mux_test.go:1080: arg len(ancestors) for printf verb %s of wrong type:
int


elithrar:
Thanks for this @mttrs!

PR #134, opened 2015-11-09
Travis: Perform gofmt, go vet checks; use race detector during tests.

shurcooL:
This change augments the Travis CI build to perform:

- Check that all files follow gofmt style, including -s (simplify) option.
- Check that go vet does not report any problems.
- Use race detector when running tests, to ensure there are no data races found.

Doing this will help catch issues like in #133 at the time the change is made, automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,14 @@
 language: go
 sudo: false
-
 go:
   - 1.3
   - 1.4
   - 1.5
   - tip
+install:
+  - go get golang.org/x/tools/cmd/vet
+script:
+  - go get -t -v ./...
+  - diff -u <(echo -n) <(gofmt -d -s .)
+  - go tool vet .
+  - go test -v -race ./...


elithrar:
LGTM. vet already passes so should not cause existing code to fail.

I should note that the race detector flag is not foolproof though as tests are usually self-contained, and therefore checking for races in tests may not actually uncover any races.

shurcooL:
> I should note that the race detector flag is not foolproof though as tests are usually self-contained, and therefore checking for races in tests may not actually uncover any races.

Agreed. I don't see it as something that provides any kind of guarantees. Instead, it's more of "given the option to run this or not, it's better to run it, just in case it helps catch issues." It's also consistent with the Travis build files I have in pretty much all of my projects - it's not specific to just gorilla/mux.

shurcooL:
Another benefit in doing this is to advertise better CI practices to more people. This is a very popular Go package, so I imagine many people will be looking at how it does things. It's better to set a better example for others to follow.

elithrar:
Yep - I am open to better Travis configs. The ones across the Gorilla
project are pretty simple (and old!) so updating these to be more
comprehensive is OK. I don't think vet (unlike lint) will require any
breaking changes to existing code either.

.travis.yml
```diff
@@ -1,8 +1,14 @@
...
 go:
   - 1.3
   - 1.4
   - 1.5
   - tip
+install:
+  - go get golang.org/x/tools/cmd/vet
+script:
+  - go get -t -v ./...
+  - diff -u <(echo -n) <(gofmt -d -s .)
```
elithrar: I assume any output from the diff will cause the test to fail?

shurcooL: That's correct.

From man pages on `diff`:

> Exit status is 0 if inputs are the same, 1 if different, 2 if trouble.

So `diff` is used precisely to cause Travis to fail if there's a non-zero `gofmt -d` output. Just running `gofmt -d` by itself has 0 exit status and wouldn't cause Travis to fail.

shurcooL:
Is there anything else that needs to be done from my side here?

elithrar:
@shurcooL - Nope, have just been busy planning a move! Thanks again for submitting this.

shurcooL:
No problem, happy to be able to help!

PR #138, opened 2015-12-23
added some methods to extract information

Snorlock:
I saw a case where I wanted to extract what I have registerd on the router so I could have a self explaining api. I wanted to have on of my routes beeing able to respond with what is publically available.

I have added a method on the Router to extract the routes registered. 
I have added some methods to be able to extract the methods that has been registered on different routers.

Please ellborate if there are better ways of doing this

### DIFF
--- a/route.go
+++ b/route.go
@@ -28,6 +28,8 @@ type Route struct {
 	strictSlash bool
 	// If true, this route never matches: it is only used to build URLs.
 	buildOnly bool
+	// If true, this route wont show in getRoutesMethod.
+	private bool
 	// The name used to build URLs.
 	name string
 	// Error resulted from building a route.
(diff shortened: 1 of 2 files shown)


route.go
```diff
@@ -73,12 +75,38 @@ func (r *Route) GetError() error {
...
+}
+
 // BuildOnly sets the route to never match: it is only used to build URLs.
 func (r *Route) BuildOnly() *Route {
 	r.buildOnly = true
 	return r
 }
 
+// MakePrivate sets the private flag making it not show up in GetRoutes.
+func (r *Route) MakePrivate() *Route {
```
elithrar: What is the goal of making a route private? I don't think this is worth adding to the public API, and it could be confused for some kind of authentication functionality.

elithrar:
I'm not sure I'm sold on this, as it's substantial addition to the size of the public API. 

Right now, if you need a way to get the routes in your application, you are better off creating a `map[string]http.Handler` and creating `Routes` from each key:value pair.

This patch effectively does the reverse, but since you control the routes you add, you shouldn't need to retrieve them from the mux.

route.go
```diff
@@ -73,12 +75,38 @@ func (r *Route) GetError() error {
...
+	r.private = true
+	return r
+}
+
+type extracter interface {
+	Match(*http.Request, *RouteMatch) bool
+	GetMatchValues() []string
+}
+
+func (r *Route) GetMethods() []string {
```
elithrar: This would need documentation if we were to merge it.

route.go
```diff
@@ -281,6 +309,10 @@ func (m methodMatcher) Match(r *http.Request, match *RouteMatch) bool {
 	return matchInArray(m, r.Method)
 }
 
+func (m methodMatcher) GetMatchValues() []string {
```
elithrar: This would need documentation if we were to merge it.

elithrar:
Going to close this one out due to a lack of activity.

Issue #20, opened 2013-04-28
Router.NotFoundHandler is not used when the router is a subrouter


ghost:
I defined specific NotFoundHandler for several subrouters but the NotFoundHandler from the main Router is always used.

error10:
I've worked around this by using a `PathPrefix()` after the home page route for the subroute.

```
    r := mux.NewRouter()
    s := r.Host("{subdomain}." + domain).Subrouter()
    //...more routes..
    s.Handle("/", commonChain.Then(subdomainHomepageHandler()))
    s.PathPrefix("/").Handler(commonChain.Then(http.HandlerFunc(subdomainNotFound)))
```

Later, `subdomainNotFound()`, which for me does some extra work, will generate a (nicely templated) 404 if necessary.

elithrar:
I'm going to close this as wontfix: if you do need separate a `NotFoundHandler` you can use nested routers (via `mux.NewRouter()`). Happy to open it again if there's a desire to.

PR #139, opened 2015-12-25
`Router.NotFoundHandler` is not used when the router is a subrouter

bign8:
Issue #20 | `Router.NotFoundHandler` is not used when the router is a subrouter
:-------|:-------
Solution | Move "Not found" logic into `Match` function.
Status | Awaiting Reviewers

This small code move should allow the "deepest" sub-router to respond with it's `NotFoundHandler`, thanks to the recursive nature of `Match`.  As this is my first pull to this repository, I expect there to be more steps to make this change than the code fix alone.  If a repository administrator leaves a comment on what else is necessary to merge to the repository, I will be happy to comply.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -59,6 +59,12 @@ func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 			return true
 		}
 	}
+
+	// Closest match for a router (includes sub-routers)
+	if r.NotFoundHandler != nil {
+		match.Handler = r.NotFoundHandler
+		return true
+	}
 	return false
 }
 
@@ -89,10 +95,7 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		setCurrentRoute(req, match.Route)
 	}
 	if handler == nil {
-		handler = r.NotFoundHandler
-		if handler == nil {
-			handler = http.NotFoundHandler()
-		}
+		handler = http.NotFoundHandler()
 	}
 	if !r.KeepContext {
 		defer context.Clear(req)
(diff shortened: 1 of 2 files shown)


elithrar:
Are you able to provide a test case for this, @bign8?

(Thanks for the PR, too)

mux_test.go
```diff
@@ -1123,6 +1123,30 @@ func TestWalkNested(t *testing.T) {
...
 
+func TestSubrouterErrorHandling(t *testing.T) {
+	superRouterCalled := false
+	subRouterCalled := false
+
+	router := NewRouter()
+	router.NotFoundHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		superRouterCalled = true
+	})
+	subRouter := router.PathPrefix("/bign8").Subrouter()
```
bign8: Can remove the self reference if desired.  Just figured I would have a little fun with the test. :wink:

bign8:
How's this @elithrar ?

mux_test.go
elithrar: Totally fine by me :+1:

elithrar:
LGTM, and thanks for adding the test. Merged!

mux.go
```diff
@@ -59,6 +59,12 @@ func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 			return true
 		}
 	}
+
+	// Closest match for a router (includes sub-routers)
+	if r.NotFoundHandler != nil {
+		match.Handler = r.NotFoundHandler
+		return true
```
alytvynov: Does it really make sense to report this as a match?
http://play.golang.org/p/SIL-WWftyQ
Seems confusing when request didn't match any real route but `router.Match` reports `true`.

FWIW I got a panic in my stats collection code which assumed that if `router.Match(...) == true`, returned `RouteMatch` would have a non-nil `Route`.

elithrar: If we don't report it as a match we would need to re-engineer how sub-router 404 handling works (e.g. revert this).

The alternative would be to set `match.Handler == nil` and then add another field to `Router` to handle this behaviour.

alytvynov: Or add a field to `RouteMatch` that says "this is a not-found handler, not an actual route match".
Or add a non-exported `matchHandler()` that does the regular `Match` behavior or returns `NotFoundHandler` if it's set.
I did not dig into the implementation of this package enough to come up with an actual valid suggestion, simply complaining :)

It basically feels like this is mixing up internal implementation details and exported API in a way that makes exported API unintuitive.

PR #152, opened 2016-03-05
Fix that `cleanPath` removes extra "/" for a URL in path

jingweno:
A path can contain a URL, e.g., /a/http://localhost/b. `path.Clean`
removes the extra "/" in "://" and makes the path look like /a/http:/localhost/b.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -10,8 +10,14 @@ import (
 	"net/http"
 	"path"
 	"regexp"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


elithrar:
I think this is better handled in https://github.com/gorilla/mux/pull/54 (i.e. optional behaviour). If you want to pick that up and address my comments there, then that'd be appreciated.

PR #155, opened 2016-03-07
Allowing case insensitive regexp groups after PR #144

jprobinson:
Changes in https://github.com/gorilla/mux/pull/144 broke regexp matching for case insensitive groups. 

This adds a test scenario and a small hunk of code reverted to what it was prior to the PR.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -370,6 +370,16 @@ func TestPath(t *testing.T) {
 			path_template: `/{product-category:a|(b/c)}/{product-name}/{product-id:[0-9]+}`,
 			shouldMatch:   true,
 		},
+		{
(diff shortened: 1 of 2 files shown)


jprobinson:
Benchmark changes: 
```
benchmark                             old ns/op     new ns/op     delta
BenchmarkMux-4                        2712          2569          -5.27%
BenchmarkMuxAlternativeInRegexp-4     3835          3860          +0.65%
BenchmarkManyPathVariables-4          4410          4476          +1.50%
```

elithrar:
Thanks @jprobinson—and thanks for adding the benchmarks (which I was just about to ask for). Happy to lose a few % if it means we maintain API compat, especially since request muxes are almost never bottlenecks in real applications.

PR #163, opened 2016-05-13
Fix golint violations

fern4lvarez:
This commit fixes:

```
mux.go:271:5: error var SkipRouter should have name of the form ErrFoo
route.go:42:1: exported method Route.SkipClean should have comment or be unexported
```

This commit does not fix violations on tests:

```
mux_test.go:35:2: don't use underscores in Go names; struct field path_template should be pathTemplate
mux_test.go:36:2: don't use underscores in Go names; struct field host_template should be hostTemplate
mux_test.go:1299:2: don't use underscores in Go names; var path_template should be pathTemplate
mux_test.go:1303:2: don't use underscores in Go names; var host_template should be hostTemplate
mux_test.go:1308:2: don't use underscores in Go names; var path_tmpl should be pathTmpl
mux_test.go:1308:13: don't use underscores in Go names; var path_err should be pathErr
mux_test.go:1313:2: don't use underscores in Go names; var host_tmpl should be hostTmpl
mux_test.go:1313:13: don't use underscores in Go names; var host_err should be hostErr
```

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -266,9 +266,9 @@ func (r *Router) Walk(walkFn WalkFunc) error {
 	return r.walk(walkFn, []*Route{})
 }
 
-// SkipRouter is used as a return value from WalkFuncs to indicate that the
+// ErrSkipRouter is used as a return value from WalkFuncs to indicate that the
 // router that walk is about to descend down to should be skipped.
-var SkipRouter = errors.New("skip this router")
+var ErrSkipRouter = errors.New("skip this router")
 
 // WalkFunc is the type of the function called for each route visited by Walk.
 // At every invocation, it is given the current route, and the current router,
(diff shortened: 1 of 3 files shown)


mux.go
```diff
@@ -266,9 +266,9 @@ func (r *Router) Walk(walkFn WalkFunc) error {
 	return r.walk(walkFn, []*Route{})
 }
 
-// SkipRouter is used as a return value from WalkFuncs to indicate that the
+// ErrSkipRouter is used as a return value from WalkFuncs to indicate that the
 // router that walk is about to descend down to should be skipped.
-var SkipRouter = errors.New("skip this router")
```
kisielk: This is part of the public API so we can't change it now.

route.go
```diff
@@ -39,6 +39,7 @@ type Route struct {
 	buildVarsFunc BuildVarsFunc
 }
 
+// SkipClean returns whether accessing "/path//to" will not redirect
```
kisielk: The phrasing of this is kind of a double negative. I'd suggest documenting it similarly to Router.SkipClean

fern4lvarez:
Since `SkipRouter` cannot be changed as it is part of the public API, the purpose of this PR is not valid anymore. Closing.

Issue #168, opened 2016-06-04
mux.Vars breaks with Go 1.7


ejholmes:
[Go 1.7 adds native support for context.Context](https://tip.golang.org/doc/go1.7) by adding [Context](https://tip.golang.org/pkg/net/http/#Request.Context) and [WithContext](https://tip.golang.org/pkg/net/http/#Request.WithContext) methods to http.Request.

The expectation is that, if you want to add anything to the context of the request, you use http.Request.WithContext, which returns a shallow copy of the request that you would pass down. Unfortunately, this seems break the way gorilla mux stores vars.

Here's a simple Go program that demonstrates the problem:

```go
package main

import (
	"context"
	"fmt"
	"net/http"
	"net/http/httptest"
	"time"

	"github.com/gorilla/mux"
)

func withTimeout(h http.Handler, timeout time.Duration) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		ctx, cancel := context.WithTimeout(r.Context(), timeout)
		defer cancel()

		r = r.WithContext(ctx)
		h.ServeHTTP(w, r)
	})
}

func ping(w http.ResponseWriter, r *http.Request) {
	fmt.Println(mux.Vars(r))
}

func main() {
	r := mux.NewRouter()
	r.Handle("/broken/{var}", withTimeout(http.HandlerFunc(ping), time.Second))
	r.Handle("/ok/{var}", http.HandlerFunc(ping))

	resp := httptest.NewRecorder()
	req, _ := http.NewRequest("GET", "/ok/foo", nil)
	r.ServeHTTP(resp, req)
	// Prints:
	// map[var:foo]

	resp = httptest.NewRecorder()
	req, _ = http.NewRequest("GET", "/broken/foo", nil)
	r.ServeHTTP(resp, req)
	// Prints:
	// map[]
}
```

Once Go 1.7 is out, basically anything that uses middleware to set context values, cancellations or deadlines is going to break mux.

It would be ideal of the vars were actually stored in the context.Context, which may be possible without breaking the API?

elithrar:
Thanks for reporting this—it does makes sense. The new `*http.Request` is a new pointer and is therefore not the same map key as the original request. I (indirectly) addressed this for gorilla/csrf - https://github.com/gorilla/csrf/pull/35 - by using build tags to use `http.Request.Context()` in >= Go 1.7.

[...]

Issue #167, opened 2016-06-02
How to set URL parameters for unit tests?


bhirbec:
Hi,

I need to set up URL parameters for my unit tests. I'm using context.Set() to do so and my test looks something like this:

```
context.Set(r, 0, map[string]string{"param1": "123"})
req , _ := http.NewRequest()
w = := httptest.NewRecorder()
myhandler(w, req)
```

The HTTP handler uses mux.Vars() to retrieve the parameters but it doesn't get any. The problem is explained in great details in [this post](http://stackoverflow.com/questions/34435185/unit-testing-for-functions-that-use-gorilla-mux-url-parameters):

> Trouble is, even when you use 0 as value to set context values, it is not same value that mux.Vars() reads. mux.Vars() is using varsKey (as you already saw) which is of type contextKey and not int.

> Sure, contextKey is defined as:

> `type contextKey int`

> which means that it has int as underlying object, but type plays part when comparing values in go, so int(0) != contextKey(0).

I'm aware I could solve my problem by either starting a web server or by calling the mux directly (as mentioned in SO). However, I don't want to do so. I'd like to call the handler directly in my test.

I see two options to solve this:
-  add a mux.SetVar() method to explicitly set URL parameters for a given request (there's already a private setVar() method)
- use `int` instead of `contextKey` type when defining `varsKey`. This way `context.Set(r, 0, map[string]string{"param1": "123"})` would be a valid way to set up URL parameters. 

Hope this help. Thanks for making Gorilla :)

Ben

elithrar:
Note: also discussed in https://github.com/gorilla/mux/pull/112

The short answer: using https://golang.org/pkg/net/http/httptest/#Server + setting `server.URL = /your/test/url` is the right way to test this. You can still test multiple handlers, use table driven tests for testing different route variables, etc - without tying your tests to mux's implementation.

The long answer: changing the type of `contextKey` to `int` would cause collisions with other packages also using the request context. The [gorilla/context docs](http://www.gorillatoolkit.org/pkg/context) cover this. Making the key type public would also add to the API and prevent us from ever changing things.

bhirbec:
@elithrar thanks for the reply. I'm not sure to understand your solution. How does the context get populated if the mux isn't involved? Can you please provide more informations or point to an example?

elithrar:
@bhirbec You populate the context by passing the correct URL - e.g.

```go
	r := mux.NewRouter()
[...]

PR #169, opened 2016-06-04
Store vars and route in context.Context when go1.7+ is used

ejholmes:
Fixes https://github.com/gorilla/mux/issues/168

This should be backwards compatible, since the usage of gorilla/context was an implementation detail and not exposed publicly.

When go1.7+ is used, the native [context.Context](https://tip.golang.org/pkg/net/http/#Request.Context) object on the http.Request will be used. On older version of go, gorilla/context will be used.


elithrar:
This LGTM! If @kisielk is happy with the PR as well we can merge.

kisielk:
I was actually thinking of doing that this weekend, but I guess you beat me to it ! 

The implementation looks great to me, very elegant.

The only thing I can think of it breaking is people who were manually extracting route vars via the gorilla/context package, but nobody should be doing that...

elithrar:
@kisielk Thankfully we didn't expose the context keys in #167 or #112!

Thanks for merging as well.

kisielk:
Yeah, that was explicitly for this reason, so we could change the context
storage method in the future.

jingweno:
:+1: Was thinking of doing this.

PR #177, opened 2016-07-25
Fix error handling in  Router.Walk

mhamrle:
In old version error was not returned properly, In one walkFn call
error was checked only for SkipRouter but not for nil.


kisielk:
I think the original assumption was that WalkFuncs would only ever return SkipRouter, but I guess it makes sense that they could also return other error values that cause an abort.

kisielk:
@elithrar any problem with merging this?

elithrar:
LGTM!

PR #185, opened 2016-08-17
Simplify extractVars, fixes edge cases.

neelance:
The current implementation was failing in a case where there are two route vars right next to each other and the second one matched the empty string. Because of that, `prevEnd < matches[i+1]` evaluated to `false` and the matches got assigned to the wrong names.

Having a capturing regexp group in a route pattern is never useful. This commit changes the tests accordingly. Another option would be to automatically turn capturing groups into non-capturing ones (adding the `?:` automatically).

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -125,12 +125,12 @@ func TestHost(t *testing.T) {
 		},
 		{
 			title:        "Host route with pattern, additional capturing group, match",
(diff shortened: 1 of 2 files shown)


elithrar:
Can you add a (pre-fix) failing test case for this?

neelance:
@elithrar I've added a test which fails before the fix.

elithrar:
Thanks for this @neelance!

PR #190, opened 2016-08-30
Add useEncodedPath option to router and routes

kushmansingh:
As discussed in #184

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -53,6 +53,8 @@ type Router struct {
 	// This has no effect when go1.7+ is used, since the context is stored
 	// on the request itself.
 	KeepContext bool
+	// see Router.UseEncodedPath(). This defines a flag for all routes.
+	useEncodedPath bool
 }
 
 // Match matches registered routes against the request.
@@ -77,7 +79,10 @@ func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 // mux.Vars(request).
 func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if !r.skipClean {
-		path := getPath(req)
+		path := req.URL.Path
+		if r.useEncodedPath {
+			path = getPath(req)
+		}
 		// Clean path to canonical form and redirect.
 		if p := cleanPath(path); p != path {
 
@@ -152,6 +157,21 @@ func (r *Router) SkipClean(value bool) *Router {
 	return r
 }
 
+// UseEncodedPath tells the router to match the encoded original path
+// to the routes.
+// For eg. "/path/foo%2Fbar/to" will match the path "/path/{var}/to".
+// This behavior has the drawback of needing to match routes against
+// r.RequestURI instead of r.URL.Path. Any modifications (such as http.StripPrefix)
+// to r.URL.Path will not affect routing when this flag is on and thus may
+// induce unintended behavior.
+//
+// If not called, the router will match the unencoded path to the routes.
+// For eg. "/path/foo%2Fbar/to" will match the path "/path/foo/bar/to"
+func (r *Router) UseEncodedPath() *Router {
+	r.useEncodedPath = true
+	return r
+}
+
 // ----------------------------------------------------------------------------
 // parentRoute
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


mux_test.go
```diff
@@ -336,8 +336,8 @@ func TestPath(t *testing.T) {
 			shouldMatch:  false,
 		},
 		{
-			title:        "Path route, URL with encoded slash does match",
-			route:        new(Route).Path("/v1/{v1}/v2"),
+			title:        "Path route with useEncodedPath, URL with encoded slash does match",
+			route:        (&Route{useEncodedPath: true}).Path("/v1/{v1}/v2"),
```
kushmansingh: I don't really like this. I'm debating whether it's worth breaking out into something like TestSkipClean

mux.go
```diff
@@ -152,6 +157,23 @@ func (r *Router) SkipClean(value bool) *Router {
...
 
+// UseEncodedPath defines the route matching behavior. The initial value is
+// false.
+//
+// When true, the router will match the encoded original path to the routes.
+// For eg. "/path/foo%2Fbar/to" will match the path "/path/{var}/to".
+// This behavior has the drawback of needing to match routes against
+// r.RequestURI instead of r.URL.Path. Any modifications (such as http.StripPrefix)
+// to r.URL.Path will not affect routing when this flag is on and thus may
+// induce uninteded behavior.
```
elithrar: typo: "unintended"

mux_test.go
elithrar: I'm all for breaking it out.

kushmansingh:
I added a separate test function for having a router with the feature turned on and also a check that it plays nicely with other features as well (as far as we can tell).

elithrar:
I'm OK with this so far.

Will merge tomorrow night PST unless any objections. Appreciate the effort
on this Kush!

markotikvic:
Hi.
I went and 'go get -u github.com/gorilla/mux' after this merge and the problem with raw path doesn't exist anymore but there seems to be a problem with UseEncodedPath method:  
```
(type *mux.Router has no field or method UseEncodedPath)  
```
But maybe I'm doing something wrong:
```
r := mux.NewRouter()
r.UseEncodedPath(true)
```
This seems OK to me, but I might be mistaken. Any ideas?

elithrar:
Not running into this myself. Is this a build-related issue?

elithrar:
Scratch that -

> I went and 'go get -u github.com/gorilla/mux' after this merge and the
problem with raw path doesn't exist anymore but there seems to be a problem
with UseEncodedPath method:

Unless you checkout this PR and `go install` it, you won't see the changes
here.

mux.go
```diff
@@ -152,6 +157,23 @@ func (r *Router) SkipClean(value bool) *Router {
...
+// When true, the router will match the encoded original path to the routes.
+// For eg. "/path/foo%2Fbar/to" will match the path "/path/{var}/to".
+// This behavior has the drawback of needing to match routes against
+// r.RequestURI instead of r.URL.Path. Any modifications (such as http.StripPrefix)
+// to r.URL.Path will not affect routing when this flag is on and thus may
+// induce unintended behavior.
+//
+// When false, the router will match the unencoded path to the routes.
+// For eg. "/path/foo%2Fbar/to" will match the path "/path/foo/bar/to"
+func (r *Router) UseEncodedPath(value bool) *Router {
```
elithrar: Better: `UseEncodedPath()`

markotikvic:
That's most likely it.  
Thank you.

mux.go
kushmansingh: Silly edge case but what if someone wants to turn it off later? They turn it on, add a couple of routes, turn it off and add some routes they don't want the behavior for.

elithrar: I don't think that'd be possible: the routes are all walked together before
serving, so you can't change the state of the *Router.

kushmansingh: Setting the flag on Router sets the flag on each of the subsequent routes, it comes into play when `route.Match()` is called.

kushmansingh: Right, that'd probably also cause some weird behavior with the regexp path variable matching. Making the change now 😄

kushmansingh:
@elithrar Any other feedback?

elithrar:
@kushmansingh None. LGTM. Merging now.

PR #196, opened 2016-09-20
document non-capturing groups (closes #143)

### LINKED ISSUE #143: Using capture groups in path regexes has unexpected behaviour
I think this is probably just a docs improvement. Unless it's buried somewhere, the docs don't specify that regexes shouldn't be wrapped in a capture group. Doing so causes unexpected behavior when used in any location but the last regex. 

Examples:

- Path: `"/{name:.*}.{ext:([a-zA-Z0-9]+)}"`
    - input `"/foo.bar"`
    - vars equals `map[name:foo ext:bar]`
- Path: `"/{name:(.*)}.{ext:[a-zA-Z0-9]+}"`
    - input `"/foo.bar"`
    - vars equals `map[name:foo ext:foo]`

It would be great if the docs clearly stated that the regexes shouldn't have an enclosing capture group. At the moment it can only be inferred from the examples, which becomes especially confusing when it works _some_ of the time (i.e. when the only capture group is the final regex).

olt:


kisielk:
Thank you!

Issue #200, opened 2016-10-12
[matching] Capture Groups Broken?


darrenmcc:
I have an endpoint that looks like this `/{type:(promo|special)}/{promoId}.json` that is now broken. `type` and `promoId` are both what was set in the `type` group.

e.g.

calling /promo/foo.json yields `map[type:promo promoId:promo]`

kallekrantz:
Hey

We're facing he same issue on our project https://github.com/thiderman/drunkenfall on master. 
Our subrouter is defined as,```/tournament/{id}/{kind:(tryout|runnerup|semi|final)}/{idx:[0-9]+}```

With the input ```/api/towerfall/tournament/lii/tryout/6/start/``` we're getting the following output ```map[id:lii kind:tryout idx:tryout]``` from ```mux.Vars```

Whereas the correct one would be ```map[id:lii kind:tryout idx:6]```

We managed to track down the revision where the issue was introduced to the following revision 0b13a922203ebdbfd236c818efcd5ed46097d690, where it's probably the simplification that is causing  the issue.

elithrar:
@darrenmcc @kallekrantz If you can submit a PR with an additional test case (e.g. your example), and that doesn't break existing tests, it'd be hugely appreciated.

kallekrantz:
Hey, @elithrar, created #202 which adds a test for it. 

Sorry if I added it to the wrong place, pretty new to Go!

elithrar:
@kallekrantz Test looks good. Did you want to also add a commit that reverts the simplification in 0b13a92 to the same PR?

(thanks heaps!)

kallekrantz:
@elithrar Sure! I'll do it during the weekend.

kallekrantz:
@elithrar reverted the commit, tests seem to be passing as well. Sorry for taking a bit longer than expected.

elithrar:
We should consider calling `panic()` if [`regexp.NumSubexp()`](https://golang.org/pkg/regexp/#Regexp.NumSubexp) != 0 - and providing a useful error message in the trace.

aeijdenberg:
Matt, I'd be happy to volunteer to write a PR including tests to do as you suggested. Would that be helpful?

elithrar:
Of course @aeijenberg - send a PR when you're ready.

aeijdenberg:
First cut of pull request added here:
https://github.com/gorilla/mux/pull/218

[...]

PR #202, opened 2016-10-13
Add test for nested regex vars

kallekrantz:
- Adds TestCorrectVars, which contains a currently failing testcase.

As requested in #200

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -1014,6 +1014,26 @@ func TestBuildVarsFunc(t *testing.T) {
 	}
 }
 
+// Addresses https://github.com/gorilla/mux/issues/200
+func TestCorrectVars(t *testing.T) {
+	tests := []routeTest{
+		{
+			title:        "Correct nested regex vars",
+			route:        new(Route).PathPrefix("/{id}/{kind:(a|b)}/{idx:[0-9]+}"),
+			request:      newRequest("GET", "http://localhost/1/a/2"),
+			vars:         map[string]string{"id": "1", "kind": "a", "idx": "2"},
+			host:         "",
+			pathTemplate: "/{id}/{kind:(a|b)}/{idx:[0-9]+}",
+			path:         "/1/a/2",
+			shouldMatch:  true,
+		},
+	}
+	for _, test := range tests {
+		testRoute(t, test)
+		testTemplate(t, test)
+	}
+}
+
 func TestSubRouter(t *testing.T) {
 	subrouter1 := new(Route).Host("{v1:[a-z]+}.google.com").Subrouter()
 	subrouter2 := new(Route).PathPrefix("/foo/{v1}").Subrouter()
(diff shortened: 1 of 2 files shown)


mux_test.go
```diff
@@ -405,16 +405,6 @@ func TestPath(t *testing.T) {
 			pathTemplate: `/{type:(?i:daily|mini|variety)}-{date:\d{4,4}-\d{2,2}-\d{2,2}}`,
 			shouldMatch:  true,
 		},
-		{
```
elithrar: Is it not possible for this test case to stay? If we keep it, do the tests fail?

kallekrantz: The test started failing if I reverted ```extractVars```. 

I tried making the test pass by modifying the function, but the testcase itself seemed a bit strange imo.

The map ```{"v1": "111", "v2": "", "v3": "222"}``` being the result of the path ```111/222``` seemed funky to me, as there seems to be route parameters missing (shouldn't match at all)

But maybe I'm just missing something

elithrar: @neelance Does failing this test break the intent of your original changes in 0b13a922203ebdbfd236c818efcd5ed46097d690?

neelance: @elithrar Yes, breaking this test means that `extractVars` is showing the faulty behavior again, which is not surprising since this PR reverts `extractVars` to its earlier implementation.

mux_test.go
```diff
@@ -1014,6 +1004,26 @@ func TestBuildVarsFunc(t *testing.T) {
 	}
 }
 
+// Addresses https://github.com/gorilla/mux/issues/200
+func TestCorrectVars(t *testing.T) {
+	tests := []routeTest{
+		{
+			title:        "Correct nested regex vars",
+			route:        new(Route).PathPrefix("/{id}/{kind:(a|b)}/{idx:[0-9]+}"),
```
neelance: This pattern has a capturing group, which is not allowed. `(a|b)` should be `(?:a|b)` instead. When using `(?:a|b)` the test passes with the `extractVars` implementation of current master. This would mean to me that #200 is invalid, because it is caused by a bad pattern input. However, some automatic check with a nicer error message would definitely be helpful in such cases.

elithrar: Yes - most users probably don't know the difference between a capturing and non-capturing group (or care), we could show an error.

However, the only path to doing so is a panic, and we'd effectively have to use a regex on the pattern first to look for non-capturing groups...

neelance: What about using a regexp replace to automatically turn capturing groups into non-capturing ones? Then we don't need a panic.

elithrar:
I'm open to that. We should also document this (in both the code, the
tests, and docs).

neelance:
Just to be clear: I don't have the time to implement this right now, quite busy with working on other projects. Sorry.

elithrar:
Indirectly addressed in https://github.com/gorilla/mux/pull/218

Issue #204, opened 2016-10-21
mux.CurrentRoute fails if context added


captncraig:
Lets say I have a middleware function that adds something to the request context:

```
func M(next http.Handler)http.Handler{
  return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request){
     r = r.WithContext(context.WithValue(r.Context(), "something",42))
     next(r,w)
  })
}
```

`mux.Handle("/foo",M(myHandler)).Name("Foo")`

if `myHandler` looks for `mux.CurrentRoute(r)`, it will get a nil result. Is there any workaround for this?

elithrar:
@captncraig Did you resolve?

PR #205, opened 2016-10-21
using context package in go 1.7+

captncraig:
I made some package variables to hold functions to get and set context values. In go 1.7+, these will be replaced with passthroughs to go's `context` package, and in prior versions, they will still use `gorilla/context` as always.

Still clear `gorilla/context` at the end of each request, so as to not break people using sessions with mux or similar use cases.

This should solve #204

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,17 @@
+// +build go1.7
+
+package mux
+
+import (
+	"context"
+	"net/http"
+)
+
+func init() {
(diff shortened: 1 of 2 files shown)


captncraig:
Tested and this does resolve #204 for me.

captncraig:
holy crap I was using an old version. nevermind me.

elithrar:
@captncraig ah, see it here. phew :)

PR #207, opened 2016-10-28
Added information to a route for documentation (Methods).

craigsammisutherland:
This patch allows adding additional information to a route. This allows for automatic generation of documentation about a route.

At the moment, the only information associated with a route is the methods.

This can be combined with the Walk() function to allow something like this:
`func documentRoute(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error {`
`template, _ := route.GetPathTemplate()`
`method, _ := route.GetInformation(mux.InformationMethods)`
`fmt.Println("[" + method + "]" + template)`
`return nil`
`}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ The name mux stands for "HTTP request multiplexer". Like the standard `http.Serv
 * [Static Files](#static-files)
 * [Registered URLs](#registered-urls)
 * [Full Example](#full-example)
+* [Retrieving Information](#retrieving-information)
 
 ---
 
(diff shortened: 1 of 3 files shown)


elithrar:
Thanks for submitting.

Can you also add a test and documentation to the README on how someone
might range over routes to get their docs?

craigsammisutherland:
Test and documentation have been added.

README.md
```diff
@@ -294,6 +295,23 @@ func main() {
...
+
+Using Walk and GetInformation it is possible to build a list of all the routes for a Router:
+
+```go
+r := mux.NewRouter()
+r.HandleFunc("/", HomeHandler).Methods("GET")
+r.HandleFunc("/products", ProductsHandler).Methods("GET", "POST")
+r.HandleFunc("/articles", ArticlesHandler).Methods("GET")
+r.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error {
+    template, _ := route.GetPathTemplate()
```
elithrar: Handle the errors in the docs.

kisielk:
I don't really like the idea of this PR. It's maintaining a secondary copy of the methods information for each route, something that doesn't seem necessary at all. If it's desirable to retrieve the list of supported methods for a route during a walk, why not add an API to do that and use the existing method list? 

Also I'm not sure about the utility of add a whole generic string "information" API, which currently has only one field that itself may be wholly redundant.

bgaifullin:
I need this patch, but I agree with @kisielk, that it is not good to store additional information about method just for documentation.
IMO: the methods can be extracted from route.matcher like this:
``` golang

func (r *Route) GetMethods() ([]string, error) {
	if r.err != nil {
		return nil, r.err
	}
        for _, m := range r.matchers {
            if methods, ok := m.(methodMatcher) {
                  return []string(methods)
            }
       }
       return nil, nil
}
```

PR #209, opened 2016-11-14
Enable to add Headers via Map

wofr:
Add headers via map

### DIFF
--- a/route.go
+++ b/route.go
@@ -218,6 +218,14 @@ func (r *Route) Headers(pairs ...string) *Route {
 	}
 	return r
 }
+// If Headers are already stored in a map
+func (r *Route) HeadersFromMap(headers map[string]string) *Route {
+	if r.err == nil {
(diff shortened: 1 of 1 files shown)


elithrar:
Can you explain the rationale behind this?

wofr:
Well I found it more "natural" adding headers from a map (key,value) rather than from a list of strings.  In my setup I got the headers I 've to use handed-over via a map and I don't not want to convert the header-map into a header-array which is than again converteded into a map.   It just about the usability.

elithrar:
HTTP headers are typically `http.Header` - which is a `map[string][]string`, not a `map[string]string`.

If you have an existing data structure, write a helper to convert. I'm not sure this adds value for all (or even most) users of this library.

wofr:
Fair enough!
Personally I find it handier handling my headers with a map rather than array of string.  This prevents me form accidently overriding header values, but of course implementing a converter works as well.

PR #212, opened 2016-12-14
Speed improvements

mekegi:
Changed O(n*m) to O(n+m) in mux.uniqueVars
and change methodMatcher and schemaMatcher from stringSlice to map [O(n) => O(1)]

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -423,13 +423,19 @@ func cleanPath(p string) string {
 
 // uniqueVars returns an error if two slices contain duplicated strings.
 func uniqueVars(s1, s2 []string) error {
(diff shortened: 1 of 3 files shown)


kisielk:
Hi,

Do you have any benchmarks to show that this is actually faster?

I'm skeptical about the schemaMatcher and methodMatcher changes, the slices are very small so the actual cost of the O(n) search should be very small.

Adding a map allocation to `uniqueVars` is going to increase the pressure on the garbage collector significantly since it's an often-used function.

kisielk:
Closing for now. Please resubmit with benchmarks if it is indeed more performant.

Issue #79, opened 2014-12-19
Subroutes using PathPrefix don't redirect on POST


DisruptiveMind:
Looking at #31 looks like it's a similar issue with POST though.

Using

``` go
bookRoute := router.PathPrefix("/book").Subrouter()
bookRoute.Handle("/", createBookHandler).Methods("POST")
```

Will only POST to **/book/** and not **/book** without the trailing slash
Trying to append **StrictSlash(true)** to the **Subrouter()** has no effect.

Is there something I'm missing in the setup?

In addition, maybe this is stupid here, but when I POST to the GET route it responds properly. So the following would both succeed, but obviously supports GET requests as well.

``` go
bookRoute := router.PathPrefix("/book").Subrouter()
get := router.Methods("GET").Subrouter()

get.Handle("/book", createBookHandler) // Handles both GET and POST
bookRoute.Handle("/", createBookHandler).Methods("POST")
```

If I remove the **bookRoute** there would only be a 404.

dcelasun:
Sorry to resurrect an old issue, but this is still relevant in 2016 and it isn't specific to `POST`. Here's an example:

```go
adminRouter := router.PathPrefix("/admin").Subrouter()
adminRouter.HandleFunc("", someHandler)
adminRouter.HandleFunc("/", someHandler)
```

With this configuration, `/admin/` works, but `/admin` does not.

elithrar:
Just to clarify, when you say "`/admin/` works, but `/admin` does not" do you mean `/admin` does not work at all and causes a 404?

dcelasun:
Exactly.

konradjurk:
@kisielk Why did you close this, it is still relevant right?

husainaloos:
@kisielk Why is this closed? I am experiencing this problem also.

eskombro:
Relevant in 2020!

regnaio:
Relevant in 2021

chancetudor:
soon to be relevant in 2022...

PR #215, opened 2017-01-16
Adding in a check for routes with just /

ShaneSaww:
This should fix https://github.com/gorilla/mux/issues/79. 
Looking for feedback!

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -1017,6 +1017,7 @@ func TestBuildVarsFunc(t *testing.T) {
 func TestSubRouter(t *testing.T) {
 	subrouter1 := new(Route).Host("{v1:[a-z]+}.google.com").Subrouter()
 	subrouter2 := new(Route).PathPrefix("/foo/{v1}").Subrouter()
+	subrouter3 := new(Route).PathPrefix("/foo").Subrouter()
 
 	tests := []routeTest{
 		{
@@ -1046,6 +1047,15 @@ func TestSubRouter(t *testing.T) {
 			host:         "",
 			path:         "/foo/bar/baz/ding",
 			pathTemplate: `/foo/{v1}/baz/{v2}`,
+			shouldMatch:  true,
+		},
+		{
+			route:        subrouter3.Path("/"),
+			request:      newRequest("GET", "http://localhost/foo/"),
+			vars:         map[string]string{},
+			host:         "",
+			path:         "/foo/",
+			pathTemplate: `/foo/`,
 			shouldMatch:  true,
 		},
 		{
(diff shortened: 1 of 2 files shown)


elithrar:
Thanks for submitting this. Will review tomorrow!

ShaneSaww:
If it helps this is the code I used to test:
```go
package main

import (
	"fmt"
	"log"
	"net/http"

	"github.com/gorilla/mux"
)

func main() {
	router := mux.NewRouter()
	adminRouter := router.PathPrefix("/admin").Subrouter()
	adminRouter.HandleFunc("", someHandler)
	adminRouter.HandleFunc("/", someHandler)

	log.Print("Listening on :5000")
	httpErr := http.ListenAndServe(":5000", router)
	if httpErr != nil {
		log.Fatal("ListenAndServe:", httpErr)
	}
}

func someHandler(w http.ResponseWriter, r *http.Request) {
	fmt.Printf("In the handler\n")
}

```

Running this should allow you to hit both 
 localhost:5000/admin
 localhost:5000/admin/

elithrar:
Can you add some additional test cases to this?

1. `/foo/bar`
2. `/foo/bar/`
3. `/foo` with a Queries() matcher

This will help prevent any regressions in future. mux has grown a lot over time & I want to make sure the testing story is as robust as it can be.

ShaneSaww:
@elithrar  Just added the extra test cases!

elithrar:
LGTM - @kisielk?

kisielk:
LGTM as well. Thanks!

scisci:
The code added doesn't seem to make any sense:

`if tpl == "/" && (len(tpl) == 0 || tpl[0] != '/') {`

Wouldn't that always evaluate to false?

ShaneSaww:
@scisci the reason that it makes sense (and it's kinda strange) is that if    

`tpl = "/"` then  `tpl[0] == 47`

 which is why  `if tpl == "/" && (len(tpl) == 0 || tpl[0] != '/') {` works. 

With all that in mind you are correct that `if len(tpl) > 0 && tpl[0] != '/' {` would make more sense.

scisci:
OK, guess I'm missing something :) 

It looked to me like this code would always return false, so it wouldn't catch truly bad templates that don't start with a slash like router.HandleFunc("someNonEmptyRouteWithNoSlash", ...)

But I'm new to gorilla so maybe this is on purpose or maybe I'm wrong.

kisielk:
I think I agree with @scisci ... looking at it again I can't seem to make this logic make sense in my head, unless I am missing something.

ShaneSaww:
After playing with it in the playground it seems broken. 
https://play.golang.org/p/0-N07DgH6W
I can push up a fix for it later today.

PR #221, opened 2017-01-25
use X-Forwarded-Host instead r.URL.Host if exist this forward header

poloskin:


elithrar:
Unfortunately we can't blindly trust this. If you need to use X-Forwarded-Host, you should put some middleware before mux's matcher that populates `r.Host` for hostnames you trust (only), or use https://godoc.org/github.com/gorilla/handlers#ProxyHeaders.

PR #224, opened 2017-02-06
Add Router.AddSubrouter() and unit tests

### LINKED ISSUE #223: Router.AddSubrouter as a new feature
A function with the signature `func (r *Router) AddSubrouter(path string, child *Router) *Router` would provide the Router-based version of the `Route.Subrouter()`.  This allows a structure along the lines of:

    users.go:
      func CreateUserEndpoint() *mux.Router {
        router := mux.NewRouter()
        ...
        return router
      }

    endpoints.go:
      type endpoint struct {
        path string
        fn   func() *mux.Router
      }

      func CreateEndpoints() *mux.Router {
        endpoints := []endpoint{
          endpoint{"/users", CreateUserEndpoint},
          ...
        }
        router := mux.NewRouter()
        for _, endp := range endpoints {
          router.AddSubrouter(endp.path, endp.fn())
        }
        return router
      }

    server.go:
      func NewRouter() *mux.Router {
        router := mux.NewRouter()
        router.AddSubrouter("/api", CreateEndpoints())
        ...
        return router
      }

I have this running in my repo - I'll submit a PR, and if this is a bad idea, feel free to close the submissions unceremoniously. :-)

tro3:
Closes #223

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -227,6 +227,37 @@ func (r *Router) HandleFunc(path string, f func(http.ResponseWriter,
 	return r.NewRoute().Path(path).HandlerFunc(f)
 }
 
+// AddSubrouter adds the given router as a subrouter underneath the
+// provided parent route.
+//
+// Similar to Route.Subrouter(), it will test the inner routes only
+// if the parent route matched. For example:
+//
+//     s := mux.NewRouter()
+//     s.HandleFunc("/products/", ProductsHandler)
+//     s.HandleFunc("/products/{key}", ProductHandler)
+//     s.HandleFunc("/articles/{category}/{id:[0-9]+}"), ArticleHandler)
+//
+//     r := mux.NewRouter()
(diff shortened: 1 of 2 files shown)


tro3:
Actually - hold off on this a bit.  Just saw some wierdness in my server behavior I need to track down and test against.

tro3:
Okay - not as simple as I thought.  The templates are created looking upward when a new Route is attached.  So if you have a Router with a bunch of Routes created _before_ that Router is added to a parent, none of the Routes below will have the proper prefixes.  This would require a recurvise-downward prefix adder function.  I may take a look at this when time allows.

This also means I misunderstood the testRoute function.  I suspect the router.Path setup was creating the path to be tested, so it masked this issue.  If I find a way to do the above, I also need to make sure the tests are nonmasking of errors.

PR #230, opened 2017-02-19
Add support for queries in URL reversing.

pamburus:
This patch allows generating an URL for a route with query parameters.

Here is an example:
```
func main() {
    router := mux.NewRouter()
    router.Path("/items/{item}").Queries("version", "{version}").Name("item")
    u, _ := router.GetRoute("item").URL("item", "123", "version", "5")
    fmt.Println(u.String())
}
```
Output:
`/items/123?version=5`

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -35,6 +35,7 @@ type routeTest struct {
 	path           string            // the expected path of the match
 	pathTemplate   string            // the expected path template to match
 	hostTemplate   string            // the expected host template to match
+	query          string            // the expected query string to match
 	shouldMatch    bool              // whether the request is expected to match the route at all
 	shouldRedirect bool              // whether the request should result in a redirect
 }
@@ -684,6 +685,7 @@ func TestQueries(t *testing.T) {
 			vars:        map[string]string{},
 			host:        "",
 			path:        "",
+			query:       "foo=bar&baz=ding",
 			shouldMatch: true,
 		},
 		{
@@ -693,6 +695,7 @@ func TestQueries(t *testing.T) {
 			vars:         map[string]string{},
 			host:         "",
 			path:         "",
+			query:        "foo=bar&baz=ding",
 			pathTemplate: `/api`,
 			hostTemplate: `www.example.com`,
 			shouldMatch:  true,
@@ -704,6 +707,7 @@ func TestQueries(t *testing.T) {
 			vars:         map[string]string{},
 			host:         "",
 			path:         "",
+			query:        "foo=bar&baz=ding",
 			pathTemplate: `/api`,
 			hostTemplate: `www.example.com`,
 			shouldMatch:  true,
@@ -724,6 +728,7 @@ func TestQueries(t *testing.T) {
 			vars:        map[string]string{"v1": "bar"},
 			host:        "",
 			path:        "",
+			query:       "foo=bar",
 			shouldMatch: true,
 		},
 		{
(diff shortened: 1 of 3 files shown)


elithrar:
Can you please explain the need behind this? An example will help.

pamburus:
Your mux package has very powerful and flexible abilities of HTTP routing.
One of them is routing by queries with ability to specify some predefined query values or use pattern-matching exactly the same way as you can do that in path routing.
But using routing by queries, it is impossible to make an URL for a some resource under that route by calling route.URL() with key/value pairs for the route variables because the resulting URL does not contain query parameters and when client uses that URL it does not return back to the same route.
Ideally, reverse routing (building URL by route) should include all parameters used from URL for routing.

If my explanation is still unclear, I can make an example.

pamburus:
I don't know how to upload go source code here, so I created an example here:
https://gist.github.com/pamburus/32d120eecacc8bc73dc77ae9174f8faa

There are two routes registered for "/items/{item}" and "/items/{item}?version={version}"
First one addresses some "item" resource without specifying specific version for the item, just referring an item with specific ID implying default version (i.e. latest).
Second one addresses some "item" resource with specific version which is specified by query parameter.

When you use first route, there are no any issues, just check:
GET http://localhost:8080/items/1
> Item: "1"
> Version: ""
> Reversed URL: "/items/1"

But when you use second route, reversed route URL does not include version, so referring to the same item without version, which is another resource, check it:
http://localhost:8080/items/1?version=20
> Item: "1"
> Version: "20"
> Reversed URL: "/items/1"

With my patch, you get the following response:
> Item: "1"
> Version: "20"
> Reversed URL: "/items/1?version=20"

Really, making an URL to the same resource which is being served by the handler (like in the example above), is not very useful. I used that just for simplicity. In real project it is often needed to refer to another resource by some set of parameters identifying the resource. 
Imagine that we have some other resource which may want to refer either to an item with some ID and default version (using first route), or to an item with specific version (using second route) and send a link to that resource in a response. It is very common for REST APIs.

elithrar:
In short: you want this feature so that you can use call `GetRoute` and have it include the query string?

It's important to provide a summary of the PR, rather than just submitting a PR. Adding more features adds complexity to the API, and we need to manage that.

(this will also need tests if we agree it is useful, before merging)

pamburus:
> In short: you want this feature so that you can use call GetRoute and have it include the query string?
Yes, more precisely, `GetRoute(name).URL(pairs...)`
> It's important to provide a summary of the PR, rather than just submitting a PR
Ok, do you have an example of such summary, or just tell what it should include.
> (this will also need tests if we agree it is useful, before merging)
I tried to find out the place where tests should be added for this feature, but there are no tests for reverse routing at all, even for paths.

kisielk:
Any PR should include a description of what the change does, and why it's necessary. If there is no description it makes it difficult to determine if we should merge the patch or not. If it's a feature addition, an example of a use case is really helpful.

Reverse routing is tested here, and this is applied to all tests which call `testRoute`:
https://github.com/gorilla/mux/blob/master/mux_test.go#L1458

It would probably be necessary to extend it to also include queries.

pamburus:
> Reverse routing is tested here, and this is applied to all tests which call testRoute:
> https://github.com/gorilla/mux/blob/master/mux_test.go#L1458

Thank you. I did not notice that. I've added description to the PR and implementing tests now. Soon I'll update the PR with the tests.

pamburus:
Tests for queries modified to test reverse routing also. Please check them.

ChrisHines:
I would like to see this functionality get merged. What work remains to be done?

elithrar:
Needs a full review (code, tests, docs, examples). You are welcome to give
it a review; I will try to make time tomorrow (Monday, PST).

As the API grows I'm enforcing more docs + examples so that godoc isn't the
only reference ;)

kisielk:
This was implemented in #270

PR #240, opened 2017-03-07
Any changes to context, etc will be preserved in the original Request

kotodharma:
Any changes to context, etc will be preserved in the original http.Request object that is passed in to ServeHTTP, for reference in later middleware processing. Without this, a new Request is spun off within the current handler, and changes made to it are lost when the scope ends.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -102,8 +102,9 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	var handler http.Handler
 	if r.Match(req, &match) {
 		handler = match.Handler
-		req = setVars(req, match.Vars)
-		req = setCurrentRoute(req, match.Route)
+		req1 := setVars(req, match.Vars)
+		req2 := setCurrentRoute(req1, match.Route)
+		*req = *req2
 	}
 	if handler == nil {
 		handler = http.NotFoundHandler()


elithrar:
Can you explain this further? unless you are concurrently modifying a request, the current approach - which saves the shallow copy - should not break applications.

olivoil:
@elithrar There's an issue at adams-sarah/test2doc#9 that this PR seems to solve

kmikiy:
Without this PR I wouldn't be able to write a middleware that runs **after** Gorilla mux handling and also use e.g. mux.Vars(r)

```go
func Middleware(next http.Handler) http.Handler {
                //...
		next.ServeHTTP(w, r)

               vars := mux.Vars(r)
               log.Printf("vars: %#+v\n", vars)
               // Will always print out map[string]string{nil}

               // but with the PR it will print out what I would expect. 
               // e.g map[string]string{"user":"201"}

               // ...
	})
}
```

PS, maybe my approach is wrong though.

kisielk:
Is this still needed now that mux has Middleware support?

elithrar:
This is now possible as of [v1.6.1](https://github.com/gorilla/mux/releases/tag/v1.6.1) as middleware is invoked after mux has performed a match, which means that route variables are available within a middleware handler.

e.g.

```go
func Logger(next http.Handler) http.Handler {
	fn := func(w http.ResponseWriter, r *http.Request) {
		vars := mux.Vars(r)
		log.Printf("url=%s method=%s vars=%s", r.URL, r.Method, vars)
		next.ServeHTTP(w, r)
	}

	return http.HandlerFunc(fn)
}
```
Output:

```sh
~ curl localhost:8080/matt
2018/01/16 14:24:48 url=/matt method=GET vars=map[name:matt]
```

PR #245, opened 2017-03-17
Router.AddRoute()

dnephin:
I'd like to build up `Route` objects before creating the router. I've run into this a couple times, and I'm wondering if you would accept this PR for an `AddRoute()` method on `Router`.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -209,9 +209,18 @@ func (r *Router) buildVars(m map[string]string) map[string]string {
 
 // NewRoute registers an empty route.
 func (r *Router) NewRoute() *Route {
-	route := &Route{parent: r, strictSlash: r.strictSlash, skipClean: r.skipClean, useEncodedPath: r.useEncodedPath}
-	r.routes = append(r.routes, route)
+	route := &Route{}
+	r.AddRoute(route)
 	return route
+}
+
+// AddRoute registers a route
+func (r *Router) AddRoute(route *Route) {
+	route.parent = r
+	route.strictSlash = r.strictSlash
+	route.skipClean = r.skipClean
+	route.useEncodedPath = r.useEncodedPath
+	r.routes = append(r.routes, route)
 }
 
 // Handle registers a new route with a matcher for the URL path.
--- a/mux_test.go
+++ b/mux_test.go
@@ -1389,6 +1389,24 @@ func TestSubrouterErrorHandling(t *testing.T) {
 	}
 }
 
+func TestAddRoute(t *testing.T) {
+	handled := false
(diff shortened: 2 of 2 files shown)


elithrar:
This would not be safe to do concurrently.

dnephin:
That's true, but that's also the case with `NewRoute()`, right?

elithrar:
Correct, but NewRoute() implies less concurrency.

Would you not create a *Router in a function and then return that instead of adding to mux's API?

e.g. 

```
package handlers

func CreateRoutes() *mux.Router {
   // create a new *Router, add routes
   return router
}
```
package main

import "handlers"

func main() {
    routes := handlers.CreateRoutes()
    // Can be applied as a Subrouter, used as the primary router, etc
}
```

Am I misunderstanding your use-case here?

**Comment by dnephin** [comment]
My use case is building up routes in different components and combining them into a single router. This is common in larger applications, or in libraries where the caller could pass in routes.

It would be possible to create a router first, and pass it around as a dependency, but as this PR demonstrates it's a somewhat artificial dependency. There's really no need to create a route from a router when you can add routes.

> `NewRoute()` implies less concurrency

I personally don't really see that implication.

**Comment by elithrar** [comment]
Can you show how NewRoute() improves things over returning a *mux.Router
from a constructor?

e.g. Show what the code looks like.

**Comment by dnephin** [comment]
Sure, here's a short example:


```go
package foo

func Routes() []*mux.Route {
    return []*mux.Route{
        mux.NewRoute()...,
        mux.NewRoute()...
    }
}
```

```go
package bar

func Routes() []*mux.Route {
    return []*mux.Route{
        mux.NewRoute()...,
        mux.NewRoute()...
    }
}
```

```go
package main

router := mux.NewRouter()
for _, route := range append(foo.Routes(), bar.Routes()...) {
    router.AddRoute(route)
}
```


The `Routes()` functions could accept a `mux.Router`, but why? It doesn't actually depend on the router to do any work. 

This is a simple case, I'm waiting for a repo to be make public and I can show something more involved.

kisielk:
> The Routes() functions could accept a mux.Router, but why? It doesn't actually depend on the router to do any work.

Isn't the point of it to add routes to the router? It should probably take a router then. If you want to defer adding of the routes you could return a function that will add routes to a router. Then you could do:

```
package main
router := mux.NewRouter()
for _, routeFn := range append(foo.Routes(), bar.Routes()...) {
    routeFn(router)
}
```

This would achieve the same thing without having to expand the API surface.

elithrar:
Timing out. Please re-open if you have further questions!

PR #251, opened 2017-04-26
Fixed golint warning

caarlos0:
Errors should be prefixed with `Err` according to golint...


elithrar:
This breaks the public API: I know it's a golint issue but was committed a long while ago. Can't merge.

PR #161, opened 2016-04-01
HTTPS Routes Building

adone:
Proposal for #13

PR #260, opened 2017-05-18
Support building URLs with non-http schemes.

### LINKED ISSUE #13: No way to build https:// URLs with URL()
As far as I can tell there's no way to generate a reverse route with an https:// URL. The only code I can find related to the scheme is in https://github.com/gorilla/mux/blob/master/route.go#L405, which sets it to the default of http, but there doesn't seem to be a way to have it generate one with https instead.

ChrisHines:
- Capture first scheme configured for a route for use when building URLs.
- Add new Route.URLScheme method similar to URLHost and URLPath.
- Update Route.URLHost and Route.URL to use the captured scheme if present.

Fixes #13. Supersedes #161.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strings"
 	"testing"
 )
@@ -31,6 +32,7 @@ type routeTest struct {
 	route          *Route            // the route being tested
 	request        *http.Request     // a request to test the route
 	vars           map[string]string // the expected vars of the match
+	scheme         string            // the expected scheme of the built URL
 	host           string            // the expected host of the built URL
 	path           string            // the expected path of the built URL
 	pathTemplate   string            // the expected path template of the route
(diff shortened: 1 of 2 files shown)


kisielk:
Looks good, but I'm not sure adding `URLScheme` is necessary. It's true that it seems to be symmetric with the rest of the API, but how useful is it really? I can't imagine it being used very often.

Also you could achieve the same thing by doing

```
schemeURL := &url.URL{Scheme: r.URL().Scheme}
```

in the case that you really need it..

ChrisHines:
Agreed. I was having doubts about `URLScheme` myself.

ChrisHines:
I'm trying to remove `URLScheme`, but it doesn't work as cleanly as I'd hope. Specifically the suggested work around of `schemeURL := &url.URL{Scheme: r.URL().Scheme}` doesn't work if the route doesn't have a host or path. Similarly, `r.URLHost()` returns an error if `r` doesn't have a configured host.

ChrisHines:
The testing is less elegant, but it works. PTAL.

ChrisHines:
Rebased on master to resolve conflicts in mux_test.go.

ChrisHines:
Somehow mux_test.go had acquired a BOM.

ChrisHines:
> Somehow mux_test.go had acquired a BOM.

The tool I used to resolve the conflict when rebasing was the source of the BOM. I've changed the settings on that so I won't have that problem again.

elithrar:
Merged. Thanks @ChrisHines!

PR #265, opened 2017-05-19
Update Walk to match all subrouters

### LINKED ISSUE #261: Walk does not walk all routes and subroutes
The current documentation on `Walk` says the following:

```
// Walk walks the router and all its sub-routers, calling walkFn for each route
// in the tree. The routes are walked in the order they were added. Sub-routers
// are explored depth-first.
```

However, this is not true: currently, `Walk` only walks routes with paths/regexps on them. This can cause `Walk` to omit subroutes that have method matchers.

Example:

```go
router := mux.NewRouter().StrictSlash(true)
products := router.PathPrefix("/{products}").Subrouter()

products.Methods("GET").Handler(...)
products.Methods("POST").Handler(...)
```

Now, call `Walk` on the router:

```go
router.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error {
	tmpl, _ := route.GetPathTemplate()
	fmt.Printf("route: %s, handler: %v\n", tmpl, route.GetHandler())
	return nil
})
```

Output:

```
route: /{products}, handler: <nil>
```

In this case, `GET` and `POST` are actually subrouters/subroutes, but are never invoked by `Walk`.

To be consistent with the documentation, these routes/routers should be visited as well. Although the `GetPathTemplate` for the routes will return `nil`/empty, it should be possible to walk the ancestors to find the nearest `Route` with a non-empty path template to determine the actual path.

This change would alter existing behavior, so it may be necessary to keep as-is and update the documentation on the function to make it clear that it is invoked only on paths (and not all routes/subroutes) and introduce a new `Walk` function for walking all of the routes.

nmiyake:
Matches all routes instead of just routes with paths.

Fixes #261

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -299,10 +299,6 @@ type WalkFunc func(route *Route, router *Router, ancestors []*Route) error
 
 func (r *Router) walk(walkFn WalkFunc, ancestors []*Route) error {
 	for _, t := range r.routes {
-		if t.regexp == nil || t.regexp.path == nil || t.regexp.path.template == "" {
-			continue
-		}
-
 		err := walkFn(t, r, ancestors)
 		if err == SkipRouter {
 			continue
--- a/mux_test.go
+++ b/mux_test.go
@@ -1329,6 +1329,34 @@ func TestWalkNested(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


elithrar:
LGTM. @kisielk?

PR #268, opened 2017-05-22
Removing half of conflict marker

brianmichel:
Looks like there was just part of a conflict marker in the README file, so I figured I'd remove it.


elithrar:
Good catch, thanks!

Issue #255, opened 2017-05-15
Route.URL() does not include query string described by Queries()


ChrisHines:
To reproduce:

```go
r := mux.NewRouter()
route := r.Path("/path/{var1}").Queries("q1", "{q1}", "q2", "{q2}")
u, _ := route.URL("var1", "1", "q1", "2", "q2", "3")
fmt.Println(u)
```

Output: `/path/1`
Expected: `/path/1?q1=2&q2=3`

ChrisHines:
It appears #230 is already in progress to address this shortcoming.

kisielk:
Implemented in #270

PR #270, opened 2017-05-30
Add support for queries in URL reversing.

### LINKED ISSUE #238: Mux not properly parsing ampersand in url query paramters
When I execute and perform a request on the following code at ```url: localhost:3000/?query=Business%20Management%20%26%20Administration```:
```
package main

import (
	"fmt"
	"io"
	"net/http"
	"github.com/gorilla/mux"
	"log"
)

func main() {
	handler := func(w http.ResponseWriter, r *http.Request) {
		fmt.Println()
		fmt.Println(r.URL.Query())
		fmt.Println(r.URL.Query().Get("query"))
		fmt.Println(r.URL.RawQuery)
		fmt.Println("Called: " + r.URL.String())
		fmt.Println("^^^")
		fmt.Println()
		io.WriteString(w, "<html><body>H World!</body></html>")
	}

	r := mux.NewRouter()
	r.HandleFunc("/", handler)
	http.Handle("/", r)
	log.Fatal(http.ListenAndServe(":3000", r))
}
```

I get the following results for my prints:
```
map[query:[Business Management ]  Administration:[]]
Business Management 
query=Business%20Management%20&%20Administration
Called: /?query=Business%20Management%20&%20Administration
^^^
```

I'm not 100% sure if it is mux, but something seems to not properly handle the ampersand. In the print for the "r.URL.RawQuery" you can see a literal "&" where a url encoded ampersand should be ("%26") and that messes up the parsing of everything else. Am I mistaken? 

I created a httptest program without mux [(here)](https://play.golang.org/p/Jhx7leJTjc) and everything came out right, so that is why I am thinking something in mux might be the cause. I apologize in advance if this is something simple that I am simply missing. Thanks for your time!

ChrisHines:
This PR builds on and supersedes #230. It fixes #238 and #255.

* Rebased the two commits by @pamburus from #230 and resolved the conflicts.
* Updated README.md and the godocs to reflect the new functionality.
* Added a test for escaped query values, which duplicated #238.
* Added a fix for #238. The fix for that—changing the default pattern for query values from `[^?&]*` to `.*`—is safe because the Go standard library has already removed the semantically important `?` and `&` characters from the query string when parsing it into individual key/value pairs. The matching logic in `routeRegexp.matchQueryString` uses the regex to match already parsed and decoded query values returned by `routeRegexp.getURLQuery`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The name mux stands for "HTTP request multiplexer". Like the standard `http.Serv
 
 * It implements the `http.Handler` interface so it is compatible with the standard `http.ServeMux`.
 * Requests can be matched based on URL host, path, path prefix, schemes, header and query values, HTTP methods or using custom matchers.
-* URL hosts and paths can have variables with an optional regular expression.
+* URL hosts, paths and query values can have variables with an optional regular expression.
 * Registered URLs can be built, or "reversed", which helps maintaining references to resources.
 * Routes can be used as subrouters: nested routes are only tested if the parent route matches. This is useful to define groups of routes that share common conditions like a host, a path prefix or other repeated attributes. As a bonus, this optimizes request matching.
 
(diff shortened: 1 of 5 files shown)


elithrar:
This is great - thank you Chris. I'll make a concerted effort to review
this ~tomorrow given the effort you put in!

kisielk:
LGTM!

Issue #279, opened 2017-06-30
route.URL() does not use scheme of parent router.


ChrisHines:
The feature added in https://github.com/gorilla/mux/pull/260 works for routes with a defined scheme, but not when the scheme is inherited from a parent router.

Reproducer:

```go
package main

import (
	"fmt"
	"github.com/gorilla/mux"
)

func main() {
	router := mux.NewRouter().Schemes("https").Host("localhost").Subrouter()
	route := router.NewRoute().Path("/api").BuildOnly()

	u, err := route.URL()

	fmt.Println(u, err)
}
```

The above code outputs `http://localhost/api <nil>`, but I was expecting `https://localhost/api <nil>`.

I have reproduced this with a new test case and will submit a PR when the implementation is working.

kisielk:
Fixed by #280

PR #280, opened 2017-06-30
Use scheme from parent router when building URLs.

ChrisHines:
Fix for #279.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -176,6 +176,13 @@ func (r *Router) UseEncodedPath() *Router {
 // parentRoute
 // ----------------------------------------------------------------------------
 
+func (r *Router) getBuildScheme() string {
+	if r.parent != nil {
+		return r.parent.getBuildScheme()
+	}
+	return ""
+}
+
 // getNamedRoutes returns the map where named routes are registered.
 func (r *Router) getNamedRoutes() map[string]*Route {
 	if r.namedRoutes == nil {
--- a/route.go
+++ b/route.go
@@ -396,7 +396,7 @@ func (r *Route) Schemes(schemes ...string) *Route {
 	for k, v := range schemes {
 		schemes[k] = strings.ToLower(v)
 	}
-	if r.buildScheme == "" && len(schemes) > 0 {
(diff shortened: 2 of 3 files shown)


mux.go
```diff
@@ -176,6 +176,13 @@ func (r *Router) UseEncodedPath() *Router {
 // parentRoute
 // ----------------------------------------------------------------------------
 
+func (r *Router) getBuildScheme() string {
+	if r.parent == nil {
```
kisielk: This is the wrong way around.

ChrisHines:
Good catch. Fixed.

route.go
```diff
@@ -396,7 +396,7 @@ func (r *Route) Schemes(schemes ...string) *Route {
 	for k, v := range schemes {
 		schemes[k] = strings.ToLower(v)
 	}
-	if r.buildScheme == "" && len(schemes) > 0 {
+	if r.getBuildScheme() == "" && len(schemes) > 0 {
```
kisielk: I'm not sure if this is correct. When `Schemes` is called it should set the schemes for the route, regardless of what the parent's schemes are. In this case the build scheme would always be the parent's, even if the route's schemes didn't include it.

ChrisHines: My thinking was that since the choice was made that building a URL uses the first scheme set, the extension of that for subrouters would be using the first scheme found starting at the root.

I am just now starting to understand the relationship between routes with schemes at multiple levels. When matching, I suppose the parent would need to include all schemes that its child routes care about. The children could either match all schemes of the parent, or a subset.

Given that relationship, I see how building a URL for a child route should prefer the scheme of the child (if present), rather than the way I implemented this.

Do you agree? If so, I can add a test to verify that behavior and update the implementation.

kisielk: Exactly. The routing will only make it to the child route if the parents also include its schemes, but the build scheme of the parent is not necessarily the one that the child will want to build with, so its schemes should take precedence if they are defined.

kisielk:
Great. Thanks a lot!

PR #281, opened 2017-07-04
Add route priority

emilevauge:
This PR adds the concept of priorities to routes.
We use gorilla/mux quite intensively on [Traefik](https://github.com/containous/traefik) and we had to add this concept to mux in order to sort routes with a specific order in the muxer.
It allows us to catch `http://mydomain.com/bigpath` before `http://mydomain.com/big` for example.
Let me know if you are not interested in this feature, I will close the PR then :)

Signed-off-by: Emile Vauge <emile@vauge.com>

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"path"
 	"regexp"
+	"sort"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


elithrar:
Priority is ascertained by the order in which routes are added, so I’m not sure this PR simplifies the API in a meaningful way. 

Further: how are identical priorities resolved? Where are the tests? 🙃

emilevauge:
@elithrar I didn't say this PR was simplifying anything.
It adds the possibility to control the priority using something else than inserted order.
I did this on a fork for Traefik because we had a use case for this.
If you this it worth it, I'll write some tests, other than that, I'll just close this PR :)

kisielk:
I think the API is a bit too specific, what if someone wants to have some other criteria?

You could always create a list of route metadata first, sort it according to priority (or whatever other method you want), and then add the routes to mux.

emilevauge:
Closing this one then.

PR #283, opened 2017-07-10
Add designation HTTP method HandleFunc()

DEATH-TROOPER:
Add designation HTTP method HandleFunc(). So, we can use HandleFunc().Methods() easily through these methods.
Ex)
`
	app.HandleFunc("/", indexHandler).Methods("GET")
`
to
`
	app.GET("/", indexHandler)
`

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -234,6 +234,23 @@ func (r *Router) HandleFunc(path string, f func(http.ResponseWriter,
 	return r.NewRoute().Path(path).HandlerFunc(f)
 }
 
+// We can use HandleFunc().Methods() easily through these.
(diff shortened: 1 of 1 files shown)


DEATH-TROOPER:
Why can't it build on Go 1.4 version. 
`
$ go tool vet .
go tool: no such tool "vet"; to install:
	go get golang.org/x/tools/cmd/vet
The command "go tool vet ." exited with 3.
`
This is error code.

elithrar:
@DEATH-TROOPER I'm not sure we want to expand the API here, as the value of these helper methods is minimal. Calling `HandleFunc` and `Methods` is slightly more verbose, but clearer.

PR #288, opened 2017-08-18
Fix #271:  Returning 405 instead of 404 when request has method mismatch with route

### LINKED ISSUE #271: 405 method not allowed vs. 404 not found
When .Method("GET") is specified and "POST" is called on that endpoint, a 404 is returned instead of a 405. The only function I can see for handling missing stuff is the NotFound() and the handler wrapper. Can we get a 405 on missing methods instead?

maknahar:
This is my attempt to handle the Issue #271 . All existing tests are passing and I have added an additional test for current feature. 

This is my first major contribution to a big open source project. Please feel free to suggest any change.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -44,7 +44,7 @@ type Router struct {
 	// Configurable Handler to be used when no route matches.
 	NotFoundHandler http.Handler
 
-	// Configurable Handler to be used request has method mismatch with route
+	// Configurable Handler to be used when the request method does not match the route.
 	MethodNotAllowedHandler http.Handler
 
 	// Parent route, if this is a subrouter.
--- a/mux.go
+++ b/mux.go
@@ -13,6 +13,10 @@ import (
 	"strings"
 )
 
+var (
+	ErrMethodMismatch = errors.New("Method is Not Allowed")
+)
+
 // NewRouter returns a new router instance.
 func NewRouter() *Router {
 	return &Router{namedRoutes: make(map[string]*Route), KeepContext: false}
@@ -106,8 +110,8 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		req = setCurrentRoute(req, match.Route)
 	}
 
-	if match.MethodMismatch {
-		handler = MethodNotAllowedHandler()
+	if handler == nil && match.MatchErr == ErrMethodMismatch {
+		handler = methodNotAllowedHandler()
 	}
 
 	if handler == nil {
@@ -351,9 +355,10 @@ type RouteMatch struct {
 	Handler http.Handler
 	Vars    map[string]string
 
-	// MethodMismatch flag is set to true if a route is matched but there
-	// is a mismatch in the request method and route method
-	MethodMismatch bool
+	// MatchErr is set to appropriate matching error
+	// It is set to ErrMethodMismatch if there is a mismatch in
+	// the request method and route method
+	MatchErr error
 }
 
 type contextKey int
@@ -556,11 +561,11 @@ func matchMapWithRegex(toCheck map[string]*regexp.Regexp, toMatch map[string][]s
 	return true
 }
 
-// MethodNotAllowed replies to the request with an HTTP status code 405.
-func MethodNotAllowed(w http.ResponseWriter, r *http.Request) {
+// methodNotAllowed replies to the request with an HTTP status code 405.
+func methodNotAllowed(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusMethodNotAllowed)
 }
 
-// MethodNotAllowedHandler returns a simple request handler
+// methodNotAllowedHandler returns a simple request handler
 // that replies to each request with a status code 405.
-func MethodNotAllowedHandler() http.Handler { return http.HandlerFunc(MethodNotAllowed) }
+func methodNotAllowedHandler() http.Handler { return http.HandlerFunc(methodNotAllowed) }
--- a/mux.go
+++ b/mux.go
@@ -43,6 +43,10 @@ func NewRouter() *Router {
 type Router struct {
 	// Configurable Handler to be used when no route matches.
 	NotFoundHandler http.Handler
+
+	// Configurable Handler to be used request has method mismatch with route
+	MethodNotAllowedHandler http.Handler
+
 	// Parent route, if this is a subrouter.
 	parent parentRoute
 	// Routes to be matched, in order.
@@ -67,6 +71,11 @@ func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 		if route.Match(req, match) {
 			return true
 		}
+	}
+
+	if match.MatchErr == ErrMethodMismatch && r.MethodNotAllowedHandler != nil {
+		match.Handler = r.MethodNotAllowedHandler
+		return true
 	}
 
 	// Closest match for a router (includes sub-routers)
--- a/mux.go
+++ b/mux.go
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	ErrMethodMismatch = errors.New("Method is Not Allowed")
+	ErrMethodMismatch = errors.New("method is not allowed")
 )
 
 // NewRouter returns a new router instance.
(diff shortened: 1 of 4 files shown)


mux.go
```diff
@@ -545,3 +555,12 @@ func matchMapWithRegex(toCheck map[string]*regexp.Regexp, toMatch map[string][]s
 	}
 	return true
 }
+
+// MethodNotAllowed replies to the request with an HTTP status code 405.
+func MethodNotAllowed(w http.ResponseWriter, r *http.Request) {
```
kisielk: I don't think this needs to be exported

mux.go
```diff
@@ -545,3 +555,12 @@ func matchMapWithRegex(toCheck map[string]*regexp.Regexp, toMatch map[string][]s
...
+	w.WriteHeader(http.StatusMethodNotAllowed)
+}
+
+// MethodNotAllowedHandler returns a simple request handler
+// that replies to each request with a status code 405.
+func MethodNotAllowedHandler() http.Handler { return http.HandlerFunc(MethodNotAllowed) }
```
kisielk: This probably doesn't need to be exported either

mux.go
```diff
@@ -344,6 +350,10 @@ type RouteMatch struct {
 	Route   *Route
 	Handler http.Handler
 	Vars    map[string]string
+
+	// MethodMismatch flag is set to true if a route is matched but there
+	// is a mismatch in the request method and route method
+	MethodMismatch bool
```
kisielk: I wonder if it would be better to have a `MatchError error` field instead. Then there could be a `ErrMethodMismatch` value used there in the case of methodMatcher, and then it could be extended with other error types in the future.

kisielk:
Overall approach looks good to me, I left some comments about possible improvements. What do you think?

maknahar:
All changes suggested by you make perfect sense @kisielk . I will make the change ASAP.

maknahar:
@kisielk I have made the changes. Please let me know if anymore improvement is required.

kisielk:
@elithrar can you give this a second look ?

mux.go
```diff
@@ -13,6 +13,10 @@ import (
 	"strings"
 )
 
+var (
+	ErrMethodMismatch = errors.New("Method is Not Allowed")
```
elithrar: Since we don't use the text for this, we should either:

* Lowercase this as convention - e.g. `method is not allowed`
* Use an enum to reflect the type of error it is.

String errors are inflexible and fragile.

mux_test.go
```diff
@@ -1871,3 +1871,29 @@ func newRequest(method, url string) *http.Request {
 	}
 	return req
 }
+
+func TestNoMatchMethodErrors(t *testing.T) {
+	func1 := func(w http.ResponseWriter, r *http.Request) {}
```
elithrar: Call it `handler`

old_test.go
```diff
@@ -121,12 +121,7 @@ func TestRouteMatchers(t *testing.T) {
 		var routeMatch RouteMatch
 		matched := router.Match(request, &routeMatch)
 		if matched != shouldMatch {
-			// Need better messages. :)
-			if matched {
-				t.Errorf("Should match.")
-			} else {
-				t.Errorf("Should not match.")
-			}
+			t.Errorf("Match Expectation: %v\nGot: %v\nRequest: %v %v", shouldMatch, matched, request.Method, url)
```
elithrar: Follow the format used elsewhere in this file - e.g. `Expected: %v Got: %v`

mux.go
```diff
@@ -344,6 +354,11 @@ type RouteMatch struct {
 	Route   *Route
 	Handler http.Handler
 	Vars    map[string]string
+
+	// MatchErr is set to appropriate matching error
+	// It is set to ErrMethodMismatch if there is a mismatch in
+	// the request method and route method
+	MatchErr error
```
elithrar: This doesn't feel right: if you provide an `error` here you can't are extremely limited in how you can customize 405 responses.

Should we not offer a `MethodNotAllowedHandler` field on the `Router` akin to `NotFoundHandler` that a package user can customize? See https://godoc.org/github.com/gorilla/mux#Router

kisielk: The purpose of this is not to customize the response, but a way for the match to signal to the router the reason why it failed to match anything. The other option is to have a flag on RouteMatch specifically for the method not matching, but that feels less extensible to me.

I agree that the router should ultimately have a way to customize the handler, but that's a separate bit of logic than this.

mux.go
maknahar: Will Change it to lower case.

mux.go
maknahar: I agree that router should have MethodNotAllowedHandler. We can default to local `methodNotAllowedHandler` if it is not set, same as we do with `http.NotFoundHandler`. Should we make those change in a current fix #271 is the question?

maknahar: @elithrar @kisielk Can we look into merging this PR if no other change is required? Or do you guys want to add `methodNotAllowedHandler` in this as well?

elithrar: @maknahar If you can add that as well, we can merge - that way it's feature-complete :)

maknahar: Sure.

maknahar:
@elithrar I have added custom `MethodNotAllowedHandler` in Router. Please have a look.

mux.go
```diff
@@ -39,6 +43,10 @@ func NewRouter() *Router {
 type Router struct {
 	// Configurable Handler to be used when no route matches.
 	NotFoundHandler http.Handler
+
+	// Configurable Handler to be used request has method mismatch with route
```
elithrar: Change to "Configurable Handler to be used when the request method does not match the route."

kisielk:
LGTM after the requested change.

maknahar:
@kisielk I have done the requested change.

kisielk:
Thanks a lot!

mux_test.go
```diff
@@ -1871,3 +1871,42 @@ func newRequest(method, url string) *http.Request {
...
+		t.Error("Should get ErrMethodMismatch error")
+	}
+
+	resp := NewRecorder()
+	r.ServeHTTP(resp, req)
+	if resp.Code != 405 {
+		t.Errorf("Expecting code %v", 405)
+	}
+
+	//Add matching route now
```
shurcooL: Minor style nit. Unlike all others, this comment is missing a space between `//` and the text.

silverweed:
This commit breaks the possibility to have a "catch-all" route, as it will always be called instead of any specific route with a different method.

Consider the following code:
```go
package main

import (
	"fmt"
	"net/http"

	"github.com/gorilla/mux"
)

func main() {
	router := mux.NewRouter()
	get := router.Methods("GET", "HEAD").Subrouter()
	post := router.Methods("POST").Subrouter()

        // Specific POST route without parameters
	post.HandleFunc("/something", func(rw http.ResponseWriter, req *http.Request) {
		fmt.Fprintln(rw, "post")
	})
        // Catch-all route that should match any GET request, and only GET requests 
	get.HandleFunc("/{anything}", func(rw http.ResponseWriter, req *http.Request) {
		fmt.Fprintln(rw, "get")
	})

	http.Handle("/", router)
	http.ListenAndServe(":8888", nil)
}
```

The expected behaviour (and the actual behaviour before this commit) is:
```sh
curl localhost:8888/something -X POST  # -> post
curl localhost:8888/foobar # -> get
```

The actual behaviour as of this commit is that both requests are answered with GET. 

This is especially problematic with routes that have the same path but different methods, as they'll all be answered with the method of the catch-all route. 

The problem disappears when lines 61-64 are commented out or removed from `route.go`.

silverweed:
Even worse:
```go
post.HandleFunc("/some/{thing}", func(rw http.ResponseWriter, req *http.Request) {
	fmt.Fprintln(rw, "post")
})
 // Same route but different method
get.HandleFunc("/some/{thing}", func(rw http.ResponseWriter, req *http.Request) {
	fmt.Fprintln(rw, "get")
})
```
```sh
curl localhost:8888/some/thing  # -> post
```

silverweed:
I [opened an issue](https://github.com/gorilla/mux/issues/300) for this.

Issue #293, opened 2017-09-08
PR for middleware support


roobre:
Hello there,

I've been using `gorilla/mux` for a while on an internal project at my organization, and due to some requirements, I ended up forking the project and adding some extra features, such as internal support for middleware.

Yes, I am aware that [negroni](https://github.com/urfave/negroni) is a thing, and that it also provides this functionality. However, there are a few key things you can't do just chaining `mux` after your middleware:

* **Acting only if a route match is found**: Waiting to see if a route actually matches before acting allows for more flexibility and coherent design. For example, an authentication middleware can happily return 403 if the required credentials are not supplied, and at the same time a normal 404 will be returned by `mux` if the route does not exist. This option does not exist if you need to process authentication headers before the request is matched.

* **Adding middleware for subrouters** is simpler if it is embed inside `mux`.

* **It is more efficient**. If your site receives heavy traffic and your middleware performs heavy tasks, you'll appreciate this kind of saving.

After pondering a bit, I decided that this simple addon was worth implementing, so I did.

As a middleware needs to be able to stop the handlers chain, I implemented it using a slightly modified `http.Handler` interface:

```go
type Middleware interface {
	ServeHTTP(http.ResponseWriter, *http.Request) (http.ResponseWriter, *http.Request)
}

type MiddlewareFunc func(http.ResponseWriter, *http.Request) (http.ResponseWriter, *http.Request)
```

If the middleware implementations return nil as either `http.ResponseWriter` or `*http.Request` the chain is stopped. Also, `http.ResponseWriter` and `*http.Request` can be hijacked if the middleware wants to. This is an approach slightly different to negroni's, where a third parameter of `*http.HandlerFunc` is added. I thought this way were better as it simplifies greatly building and iterating the middleware chain. Also, I don't like recursion. Of course, wrapper functions which receive `http.Handler` and `http.HandlerFunc` exists, so standard handlers can be chained too.

I'm not sure if you think this feature is kind of out-of scope. But if you don't, I'll happily open a PR and discuss any modifications or style changes you'd want to be done.

Abdullah2993:
I second this proposal. it will actually be very useful...I'm currently working on an authentication middleware so I wanted to add the ability to allow certain routes without authentication but can't because `mux.CurrentRoute` returns nil if not called from the the handler it self

Abdullah2993:
Can I send a pull request with something like this
https://github.com/gorilla/mux/compare/master...Abdullah2993:master

[...]

PR #294, opened 2017-09-11
Add middleware support as discussed in #293

roobre:
See discussion in #293 for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,8 +450,10 @@ func main() {
 
 ### Middleware
 
-Since **vX.Y.Z**, mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed if a
-match is found (including subrouters). Middlewares are defined using the de facto standard type:
+Mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed in the order they are added if a match is found (including subrouters).
+Middlewares are (typically) small pieces of code which take one request, do something with it (e.g. log some fields, add/remove some headers), and pass it down to another middleware or the final handler.
+
+Mux middlewares are defined using the de facto standard type:
 
 ```go
 type MiddlewareFunc func(http.Handler) http.Handler
@@ -504,8 +506,10 @@ func (amw *authenticationMiddleware) Middleware(next http.Handler) http.Handler 
         if user, found := amw.tokenUsers[token]; found {
         	// We found the token in our map
         	log.Printf("Authenticated user %s\n", user)
+        	// Pass down the request to the next middleware (or final handler)
         	next.ServeHTTP(w, r)
         } else {
+        	// Write an error and stop the handler chain
         	http.Error(w, "Forbidden", 403)
         }
     })
@@ -522,7 +526,7 @@ amw.Populate()
 r.AddMiddlewareFunc(amw.Middleware)
 ```
 
-Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to.
+Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to. Middlewares *should* write to `ResponseWriter` if they *are* going to terminate the request, and they *should not* write to `ResponseWriter` if they *are not* going to terminate it.
 
 ## Full Example
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ The name mux stands for "HTTP request multiplexer". Like the standard `http.Serv
 * [Registered URLs](#registered-urls)
 * [Walking Routes](#walking-routes)
 * [Graceful Shutdown](#graceful-shutdown)
+* [Middleware](#middleware)
 * [Full Example](#full-example)
 
 ---
@@ -446,6 +447,82 @@ func main() {
     os.Exit(0)
 }
 ```
+
+### Middleware
+
+Since **vX.Y.Z**, mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed if a
+match is found (including subrouters). Middlewares are defined using the de facto standard type:
+
+```go
+type MiddlewareFunc func(http.Handler) http.Handler
+```
+
+Typically, the returned handler is a closure which does something with the http.ResponseWriter and http.Request passed to it, and then calls the handler passed as parameter to the MiddlewareFunc (closures can access variables from the context where they are created).
+
+A very basic middleware which logs the URI of the request being handled could be written as:
+
+```go
+func simpleMw(next http.Handler) http.Handler {
+    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+        // Do stuff here
+        log.Println(r.RequestURI)
+        // Call the next handler, which can be another middleware in the chain, or the final handler.
+        next.ServeHTTP(w, r)
+    })
+}
+```
+
+Middlewares can be added to a router using `Router.AddMiddlewareFunc()`:
+
+```go
+r := mux.NewRouter()
+r.HandleFunc("/", handler)
+r.AddMiddleware(simpleMw)
+```
+
+A more complex authentication middleware, which maps session token to users, could be written as:
+
+```go
+// Define our struct
+type authenticationMiddleware struct {
+	tokenUsers map[string]string
+}
+
+// Initialize it somewhere
+func (amw *authenticationMiddleware) Populate() {
+	amw.tokenUsers["00000000"] = "user0"
+	amw.tokenUsers["aaaaaaaa"] = "userA"
+	amw.tokenUsers["05f717e5"] = "randomUser"
+	amw.tokenUsers["deadbeef"] = "user0"
+}
+
+// Middleware function, which will be called for each request
+func (amw *authenticationMiddleware) Middleware(next http.Handler) http.Handler {
+    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+        token := r.Header.Get("X-Session-Token")
+        
+        if user, found := amw.tokenUsers[token]; found {
+        	// We found the token in our map
+        	log.Printf("Authenticated user %s\n", user)
+        	next.ServeHTTP(w, r)
+        } else {
+        	http.Error(w, "Forbidden", 403)
+        }
+    })
+}
+```
+
+```go
+r := mux.NewRouter()
+r.HandleFunc("/", handler)
+
+amw := authenticationMiddleware{}
+amw.Populate()
+
+r.AddMiddlewareFunc(amw.Middleware)
+```
+
+Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to.
 
 ## Full Example
 
--- a/README.md
+++ b/README.md
@@ -450,8 +450,8 @@ func main() {
 
 ### Middleware
 
-Mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed in the order they are added if a match is found (including subrouters).
-Middlewares are (typically) small pieces of code which take one request, do something with it (e.g. log some fields, add/remove some headers), and pass it down to another middleware or the final handler.
+Mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed in the order they are added if a match is found, including its subrouters.
+Middlewares are (typically) small pieces of code which take one request, do something with it, and pass it down to another middleware or the final handler. Some common use cases for middleware are request logging, header manipulation, or `ResponseWriter` hijacking.
 
 Mux middlewares are defined using the de facto standard type:
 
@@ -459,7 +459,7 @@ Mux middlewares are defined using the de facto standard type:
 type MiddlewareFunc func(http.Handler) http.Handler
 ```
 
-Typically, the returned handler is a closure which does something with the http.ResponseWriter and http.Request passed to it, and then calls the handler passed as parameter to the MiddlewareFunc (closures can access variables from the context where they are created).
+Typically, the returned handler is a closure which does something with the http.ResponseWriter and http.Request passed to it, and then calls the handler passed as parameter to the MiddlewareFunc. This takes advantage of closures being able access variables from the context where they are created, while retaining the signature enforced by the receivers.
 
 A very basic middleware which logs the URI of the request being handled could be written as:
 
--- a/doc.go
+++ b/doc.go
@@ -238,5 +238,70 @@ as well:
 	url, err := r.Get("article").URL("subdomain", "news",
 	                                 "category", "technology",
 	                                 "id", "42")
+
+Since **vX.Y.Z**, mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed if a
+match is found (including subrouters). Middlewares are defined using the de facto standard type:
+
+	type MiddlewareFunc func(http.Handler) http.Handler
+
+Typically, the returned handler is a closure which does something with the http.ResponseWriter and http.Request passed to it, and then calls the handler passed as parameter to the MiddlewareFunc (closures can access variables from the context where they are created).
+
+A very basic middleware which logs the URI of the request being handled could be written as:
+
+	func simpleMw(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Do stuff here
+			log.Println(r.RequestURI)
+			// Call the next handler, which can be another middleware in the chain, or the final handler.
+			next.ServeHTTP(w, r)
+		})
+	}
+
+Middlewares can be added to a router using `Router.AddMiddlewareFunc()`:
+
+	r := mux.NewRouter()
+	r.HandleFunc("/", handler)
+	r.AddMiddleware(simpleMw)
+
+A more complex authentication middleware, which maps session token to users, could be written as:
+
+	// Define our struct
+	type authenticationMiddleware struct {
+		tokenUsers map[string]string
+	}
+
+	// Initialize it somewhere
+	func (amw *authenticationMiddleware) Populate() {
+		amw.tokenUsers["00000000"] = "user0"
+		amw.tokenUsers["aaaaaaaa"] = "userA"
+		amw.tokenUsers["05f717e5"] = "randomUser"
+		amw.tokenUsers["deadbeef"] = "user0"
+	}
+
+	// Middleware function, which will be called for each request
+	func (amw *authenticationMiddleware) Middleware(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			token := r.Header.Get("X-Session-Token")
+
+			if user, found := amw.tokenUsers[token]; found {
+				// We found the token in our map
+				log.Printf("Authenticated user %s\n", user)
+				next.ServeHTTP(w, r)
+			} else {
+				http.Error(w, "Forbidden", 403)
+			}
+		})
+	}
+
+	r := mux.NewRouter()
+	r.HandleFunc("/", handler)
+
+	amw := authenticationMiddleware{}
+	amw.Populate()
+
+	r.AddMiddlewareFunc(amw.Middleware)
+
+Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to.
+
 */
 package mux
--- a/doc.go
+++ b/doc.go
@@ -257,7 +257,7 @@ A very basic middleware which logs the URI of the request being handled could be
 		})
 	}
 
-Middlewares can be added to a router using `Router.AddMiddlewareFunc()`:
+Middlewares can be added to a router using `Router.Use()`:
 
 	r := mux.NewRouter()
 	r.HandleFunc("/", handler)
@@ -299,7 +299,7 @@ A more complex authentication middleware, which maps session token to users, cou
 	amw := authenticationMiddleware{}
 	amw.Populate()
 
-	r.AddMiddlewareFunc(amw.Middleware)
+	r.Use(amw.Middleware)
 
 Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to.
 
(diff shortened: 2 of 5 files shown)


roobre:
Status update: I've rebased `middleware` branch to match latest version of `master`. Now middlewares are not executed if a not found or method mismatch event occurs, even if custom handlers are used. I've added test cases for these, and improved some existing ones.

As this commit list is getting huge, I'll squash them after you review individual changes.

README.md
```diff
@@ -383,6 +383,41 @@ r.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error
 })
 ```
 
+### Middleware
+
+
+Since **vX.Y.Z**, mux supports the addition of middlewares to a router, which are executed if a
```
elithrar: Replace "router" with `[Router](https://godoc.org/github.com/gorilla/mux#Router)` so that we link to the docs.

README.md
```diff
@@ -383,6 +383,41 @@ r.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error
...
+match is found (including subrouters). Middlewares are defined using the de facto standard function:
```
elithrar: s/function/type

README.md
```diff
@@ -383,6 +383,41 @@ r.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error
...
+
+```go
+func(http.Handler) http.Handler
+```
+
+Structs that implement the `Middleware` interface can also be used:
```
elithrar: s/Structs/Types

README.md
```diff
@@ -383,6 +383,41 @@ r.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error
...
+
+```go
+type Middleware interface {
+    Middleware(handler http.Handler) http.Handler
+}
+```
+
+Middlewares can be added to a router using `Router.AddMiddlewareFunc()` and
```
elithrar: Examples for both please!

README.md
```diff
@@ -383,6 +383,41 @@ r.Walk(func(route *mux.Route, router *mux.Router, ancestors []*mux.Route) error
...
+match is found (including subrouters). Middlewares are defined using the de facto standard function:
+
+```go
+func(http.Handler) http.Handler
+```
+
+Structs that implement the `Middleware` interface can also be used:
+
+```go
+type Middleware interface {
```
elithrar: I'm not entirely convinced this interface is useful. Can you expand more on how this is a better API vs. only supporting `func(http.Handler) http.Handler` - ?

roobre:
Thanks for the comments. You're right about the docs, its a bit lacking. `Middleware` interface is useful for building complex middlewares which need to hold some data and/or extra logic. Also it gives the API some parallelism with `http.Handler` and `http.HandlerFunc`: Same concept, but in different places. Other routers and middleware managers use the same strategy, as it is a simple way to keep complex middlewares tidy and clean.

I'll add extra examples and fix the issues you pointed out over the course of this week, I'll be pretty busy at work until Wed :(

roobre:
I've fixed some of the issues you pointed out in the commit above. Please tell me if there are something else which needs improvement.

elithrar:
>  Also it gives the API some parallelism with http.Handler and http.HandlerFunc: Same concept, but in different places. Other routers and middleware managers use the same strategy, as it is a simple way to keep complex middlewares tidy and clean.

Do you have examples? I'm partial to [chi](https://github.com/go-chi/chi), which has a `Use((http.Handler) http.Handler)` method on the router, but the concept of a Middleware interface seems less common.

For context: adding two method is 100% more decision-making that a package user has to make, and adds to our already extensive package API. Thus, I'm pushing hard for utmost simplicity unless there's a clear case where the interface-based support makes sense. I can't see that it makes sense right now.

roobre:
There is an example on https://github.com/roobre/gorilla-mux/blob/middleware/README.md#middleware, line 427:

```go
// Define our struct
type authenticationMiddleware struct {
	tokenUsers map[string]string
}

// Initialize it somewhere
func (amw *authenticationMiddleware) Populate() {
	amw.tokenUsers["00000000"] = "user0"
	amw.tokenUsers["aaaaaaaa"] = "userA"
	amw.tokenUsers["05f717e5"] = "randomUser"
	amw.tokenUsers["deadbeef"] = "user0"
}

// Middleware function, which will be called for each request
func (amw *authenticationMiddleware) Middleware(next http.Handler) http.Handler {
        token := r.Header.Get("X-Session-Token")
        
        if user, found := amw.tokenUsers[token]; found {
        	// We found the token in our map
        	log.Printf("Authenticated user %s\n", user)
        	next.ServeHTTP(w, r)
        } else {
        	http.Error(w, "Forbidden", 403)
        }
    })
}
```

```go
r := mux.NewRouter()
r.HandleFunc("/", handler)

amw := authenticationMiddleware{}
amw.Populate()

r.AddMiddleware(&amw)
```

Without implementing this interface, instead of `mux.AddMiddleware(authenticator)` we would have to write `mux.AddMiddlewareFunc(authenticator.Middleware)`, which seems a bit clunkier to me.

It's just the same as `http.Handler`. Go devs could have left the API without the `Handler` interface and everything would work the same, but instead of `http.ListenAndServe(":8000", mux)` we would have to write `http.ListenAndServe(":8000", mux.ServeHTTP)`.

The extra decision making is indeed here, but it is for a reason: You either want a simple middleware which, for example, logs requests, or a complex one which can hold data structures, have extra logic, etc. Just like it happens with handlers.

The `Middleware` interface helps code reusability too, as having a middleware interface kind of forces users to name their middleware methods all the same, so they can be exchanged with ease. Let try to give an example: Let's suppose my logger middleware becomes extremely popular, and I decide to give it a separate package: `roob.re/loggermiddleware`. A potential user which already have another middleware, let's say, `randomRejectMiddleware`, wants to add it to their stack:

```go
package main

import (
    "net/http"
    "github.com/gorilla/mux"
    "roob.re/loggermiddleware"
)

// [...]

func main() {
    mux := mux.NewRouter()

    rrmw := &randomRejectMiddleware{}
    lgmw := loggermiddleware.New(log)

    mux.AddMiddleware(rrmw)
    mux.AddMiddleware(lgmw)

    // [...]
}
```

This looks simple enough. Without the middleware interface, we could have called our `Middleware(http.Handler) http.Handler` function different, and it would probably look like:


```go
package main

import (
    "net/http"
    "github.com/gorilla/mux"
    "roob.re/loggermiddleware"
)

// [...]

func main() {
    mux := mux.NewRouter()

    rrmw := &randomRejectMiddleware{}
    lgmw := loggermiddleware.New(log)

    mux.AddMiddleware(rrmw.DoMiddleware)
    mux.AddMiddleware(lgmw.MiddlewareLogic)

    // [...]
}
```

I know the names are quite absurd, but you get the point. The middleware interface forces an standard middleware writes can adhere to, and makes the code to add them look simpler and nicer. Just like `http.Handler` does.

advanderveer:
Any progress on this? Would love to use this for a new project ;)

elithrar:
Still needs review, and I’m just getting back from vacation. Soon!

xgfone:
I like this feature, too!

Abdullah2993:
any status update so far?

middleware.go
```diff
@@ -0,0 +1,21 @@
+package mux
+
+import "net/http"
+
+type Middleware interface {
```
elithrar: I don't think we should keep this. Single method interfaces can be restrictive; best to let the package user define a wrapper if needed. We end up adding similarly-named methods to the API.

(if there is demand over time, we can re-assess)

doc.go
```diff
@@ -238,5 +238,31 @@ as well:
...
+A very basic middleware which logs the URI of the request being handled could be written as:
+
+	func simpleMw(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			log.Println(r.RequestURI)
+			next.ServeHTTP(w, r)
+		})
+	}
+
+Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()`
```
elithrar: We should document/clarify how handlers are executed (in order they added).

elithrar:
@kisielk - let me know if you echo my thoughts here.

doc.go
roobre: Roger, I'll clarify the calling order and stopping condition.

middleware.go
roobre: I still think that adding the `Middleware` interface will help to keep middlewares more standard, and the parallelism with `net/http` `Handler` and `HandlerFunc` helps users to understand the behaviour. This pattern is also seen in other routers and middleware managers (see negroni, for example).

However, if you think it doesn't fit the project, I'll remove it.

roobre:
I'll address the changes listed above when I return from vacation. I'm sorry for the inconvenience :(

elithrar:
No inconvenience at all!

elithrar:
@roobre - let me know if you need any help here. Would love to help get this merged in! 🙌🏼

roobre:
Sorry, I'm on vacation until maybe 3-4 Jan, so I've been quite inactive here :_

I'll get this rebased and fixed as soon as I get home. But I'd like to see if you have reached a consensus about the `Middleware` interface (basically if I should remove it or not /cc @kisielk)

kisielk:
Hm, I'm not sure how I feel about the interface. What would the alternative look like?

elithrar:
@kisielk 

Without the interface, we should just have (as the public API):

```
// Takes a func (http.Handler) http.Handler
route.AddMiddleware(mw)
```
The interface we rally around is `http.Handler` rather than anything else.

There may be *some desire for a `http.HandlerFunc` option - 

```
// Takes a func (http.HandlerFunc) http.HandlerFunc
router.AddMiddlewareFunc(mw)
```

... but I'd rather tease that out by starting minimally.

kisielk:
Ah I understand now. Yes I think just using `func (http.Handler) http.Handler` as a type for the middleware is just fine. No need to make everything into an interface that you need to implement. A much less "object-oriented" approach and more a functional transformation.

roobre:
@kisielk It's not exactly as you said. The interface just wraps `func (http.Handler) http.Handler` as a receiver and gives it a name. Just like `http.Handler` does with `http.HandlerFunc`. You can see a detailed explanation above: https://github.com/gorilla/mux/pull/294#issuecomment-343438301

kisielk:
Yeah, I understand that, but having the `MiddleWare` method on the type is unnecessary.

```
// Define our struct
var tokenUsers :=  map[string]string{
	"00000000": "user0",
	"aaaaaaaa": "userA",
	"05f717e5": "randomUser",
	"deadbeef": "user0",
}

// Middleware function, which will be called for each request
func AwmMiddleware(next http.Handler) http.Handler {
        token := r.Header.Get("X-Session-Token")
        
        if user, found := tokenUsers[token]; found {
        	// We found the token in our map
        	log.Printf("Authenticated user %s\n", user)
        	next.ServeHTTP(w, r)
        } else {
        	http.Error(w, "Forbidden", 403)
        }
    })
}
```

Then you don't need to create additional type boilerplate. At this point the interface would be created solely for the purpose of calling one method on Mux, which seems like it's unnecessary.

ckeyer:
like

roobre:
It's there just to standarize the receiver name, just like `http.Handler`. They're even defined the same way! (see https://golang.org/pkg/net/http/#Handler and https://golang.org/pkg/net/http/#HandlerFunc).

Anyway, in the end it's not my decision to make, so I'll rebase, squash, and remove it. Please ping me if you change your mind.

roobre:
Ok, I've made the `Middleware` interface private. This is because if, in the future, we change our mind and decide to make it public, it will be a lot easier than if I just wipe it from existance:

* The middleware list would no longer be of type `middleware`, but a function instead, which would require a dirty rewrite of the `addMiddleware` method (we can easily store `MiddlewareFunc`s as `middleware`s because `MiddlewareFunc` implements `middleware`, but the opposite would be dirty).

* Having a type named `MiddlewareFunc` instead of just `Middleware` would be a bit confusing, and certainly problematic if in the future we wanted to go back to both APIs.

Please tell me what you think about this. As always, let me kindly remind you that this is, from my humble point of view, a mistake. The "Interface plus function which implements the interface" is a common pattern in go, and the idiomatic way of doing things, as we can clearly see in the net/http design and implementation.

elithrar:
Thanks for the update.

I do understand your consternation here, but since we already have a widely
used function signature in the wild + in-use elsewhere, pragmatism beats
out purism.

If this were “greenfields”, as it were, defining our own interface would
make more sense 😉

elithrar:
Plus, as you say: we can expose the interface (and the extra methods)
without too much pain if there’s demand.

mux has a big API and thus the friction to add to it should grow over time
in order to keep it digestible to users.

roobre:
I understand your points too, and I'm happy with the middleware interface staying, but private :D

I'll change the pertinent parts of the docs and comments and squash tomorrow evening (GMT).

PD: Btw, it looks like Travis hasn't complained for this PR, but the instance I keep on my repo yelled at me for some vet things on 1.6 (???) https://travis-ci.org/roobre/gorilla-mux/jobs/325103855
I guess I shouldn't worry about this, but I'll leave it here just in case.

elithrar:
Thanks for noting RE: vet. I couldn't see any relevant changes in the Go 1.6 release notes, but we'll see if it comes up again. I suspect it might be trying to overly-aggressively check whether `Route` exists as a top-level type (which it does not).

roobre:
Rebased to latest commits on `master`.

elithrar:
Good to review?

README.md
```diff
@@ -447,6 +448,82 @@ func main() {
 }
 ```
 
+### Middleware
+
```
elithrar: * Add a quick paragraph describing what middleware is / what it can be used for, and link to https://github.com/gorilla/handlers
* Version will be v1.6.1 if we merge this week (I'll cut a new release after this PR is merged in)

e.g.

> Middleware describes intermediate handlers that can check, mutate or log the incoming request or outgoing response. mux makes it easy to add middleware to a [Router](https://godoc.org/github.com/gorilla/mux#Router) or Subrouter...

README.md
```diff
@@ -447,6 +448,82 @@ func main() {
...
+r := mux.NewRouter()
+r.HandleFunc("/", handler)
+
+amw := authenticationMiddleware{}
+amw.Populate()
+
+r.AddMiddlewareFunc(amw.Middleware)
+```
+
+Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to.
```
elithrar: > "This can be used to avoid passing the request on, but make sure to return a response to the client from within the middleware"

> Note: a library of useful logging, CORS and proxy-related middleware can be found in the [gorilla/handlers](https://github.com/gorilla/handlers) package

doc.go
```diff
@@ -238,5 +238,70 @@ as well:
 	url, err := r.Get("article").URL("subdomain", "news",
 	                                 "category", "technology",
 	                                 "id", "42")
+
+Since **vX.Y.Z**, mux supports the addition of middlewares to a [Router](https://godoc.org/github.com/gorilla/mux#Router), which are executed if a
```
elithrar: Update to align with the README

(Note to self: use tooling to make this automatic...)

middleware.go
```diff
@@ -0,0 +1,28 @@
...
+// Typically, the returned handler is a closure which does something with the http.ResponseWriter and http.Request passed
+// to it, and then calls the handler passed as parameter to the MiddlewareFunc.
+type MiddlewareFunc func(http.Handler) http.Handler
+
+// middleware interface is anything which implements a MiddlewareFunc named Middleware.
+type middleware interface {
+	Middleware(handler http.Handler) http.Handler
+}
+
+// MiddlewareFunc also implements the Middleware interface.
```
elithrar: Update this? As we don't export the `Middleware` interface (`middleware`) this may be confusing.

middleware.go
```diff
@@ -0,0 +1,28 @@
...
+func (mw MiddlewareFunc) Middleware(handler http.Handler) http.Handler {
+	return mw(handler)
+}
+
+// AddMiddlewareFunc appends a MiddlewareFunc to the chain.
+func (r *Router) AddMiddlewareFunc(mwf MiddlewareFunc) {
```
elithrar: `AddMiddleware`, `Add` or `UseMiddleware` or just `Use` might be less verbose here.

* `chi` and `echo` call their methods `Use`
* [express](http://expressjs.com/en/guide/using-middleware.html) also calls it `use`
* Downside to `Use` - not entirely self-documenting.

I'd be OK with just calling this function `Middleware` as "Add" is obvious, and subtracting it/removing it later is (mostly) a non-option.

middleware.go
```diff
@@ -0,0 +1,28 @@
...
+// middleware interface is anything which implements a MiddlewareFunc named Middleware.
+type middleware interface {
+	Middleware(handler http.Handler) http.Handler
+}
+
+// MiddlewareFunc also implements the Middleware interface.
+func (mw MiddlewareFunc) Middleware(handler http.Handler) http.Handler {
+	return mw(handler)
+}
+
```
elithrar: > `Middleware` adds a middleware to the chain. Middleware can be used to intercept or otherwise modify requests and/or responses, and are executed in the order that they are applied to the Router.

middleware.go
roobre: Do you mean the comment or the implementation?

The implementation needs to be there in order to store `MiddlewareFunc` as `Middleware` in the `middlewares` slice. But the comment needs to be updated indeed, I slipped there :P

middleware.go
roobre: I don't like the `AddMiddlewareFunc` name either, seems way too long to me. But I think we should keep the "func" part somewhere, as it makes what this function expects very clear, and avoids the need for a breaking change if the `middleware` interface is exported in the future.

What about `UseFunc`?

roobre:
I'll patch this along the weekend, probably tomorrow.

middleware.go
elithrar: Just the comment 👌🏼

middleware.go
elithrar: Let’s just stick with `Use` - as that’s familiar and aligns with the `Handle` (takes a http.Handler) and `HandleFunc` (takes a HandlerFunc) on `Router`. This somewhat sits in the middle, but familiarity wins out in lieu of other objections.

README.md
roobre: Added!

README.md
roobre: Added, but reworded "make sure" to "should", as some particular scenarios may apply and it is not a mux requirement.

doc.go
roobre: I removed the version thingie. I'm not sure if its an important enough information to be in the readme. Feel free to re-add it if you think it is.

middleware.go
roobre: Done!

middleware.go
roobre: Done. I've renamed the internal `addMiddleware` to `useInterface`, to avoid collision with `Use`.

roobre:
Fixed most of the things, please tell me what you think :D

doc.go
elithrar: Good move. I think it's fair to assume that the documentation reflects `HEAD`.

elithrar:
OK - I'm going to test this in an application I have running (hopefully tonight) so we can get this merged.

Excited to land this.

@kisielk - if you can review in the interim that'd be great!

roobre:
There you go, @kisielk. I appreciate feedback on these grammar mistakes, as I'm not a native speaker I sometimes slip things here and there :(

kisielk:
That one is just something that stands out because it used to be a really bad habit of mine. I used to write stuff that had parentheses in nearly every sentence because that's just how my train of thought works. Then based on feedback I realized it doesn't necessarily make for the easiest reading for other people :)

README.md
roobre: Btw, I forgot to mention I did not add the link because most handler don't actually implement `mux.Middleware`, as they require additional parameters. We could circunvent this using closures, but I think it would confuse some users instead of adding clarity. Maybe we could rework some handlers to make them comply with the interface?

kisielk: Maybe instead of changing them we could add a `XXXXMiddleware` function for each one?

roobre:
Same here, I usually think the main points while mentally adding small exceptions and clarifications here and there, so I tend to add parenthesis and clarifications everywhere instead of rewording the whole thingk and making it more linear. Anyway, as I said, I appreciate the feedback :D

elithrar:
OK: sanity checked locally and all works as expected, esp. around Subrouters. I'm going to put this in prod soon to expose some heap & goroutine profiles behind authentication, so timing is good!

elithrar:
Thanks for your work on this @roobre—appreciate your responsiveness and patience!

elithrar:
Tagged in v1.6.1 - https://github.com/gorilla/mux/releases/tag/v1.6.1

README.md
roobre: Yes, of course. We should not break existing handlers, so maybe wrapping them somehow, or refactoring the core logic into a different, private function and exposing both the actual and the Middleware flavours of it. I can give it a shot if you think it's worth it.

iambudi:
Trying to use middleware but i have this error message.
`r.AddMiddleware undefined (type *mux.Router has no field or method AddMiddleware)`

```
r := mux.NewRouter()
r.HandleFunc("/", handler)
r.AddMiddleware(myMiddleware)
```

kisielk:
It should be `r.Use(myMiddleware)`. Seems we forgot to adjust that one example in the docs.

kisielk:
Updated the docs in  85e6bff, @elithrar should we tag a new release for the doc update?

iambudi:
@kisielk Thanks. It works now.

elithrar:
See https://github.com/gorilla/mux/pull/332 for the docs fix.

doc.go
dgellow: Hi there. I can see that the online documentation still has the `*vX.Y.Z*`. Also the link `[Router](...)` hasn't been converted.

screen shot 2018-01-18 at 11 58 40

roobre: Yikes, I totally missed that one :/
I can fix it and create a new PR if you want, but maybe you can edit it quicker @elithrar

elithrar: I’ll take a look today. The website doesn’t do Markdown, which is a bit of a pain.

elithrar: https://github.com/gorilla/mux/pull/333

dgellow: Cool, thanks guys :)

kgadams:
Hi,
I love this feature, however I think you should clarify in the documentation if the middelware needs to be concurrency safe (I assume it does).
So the authenticationMiddleware only works if the amw.tokenUsers hash is never updated while the webserver is running. Otherwise you get nasty races.

roobre:
Middleware does indeed need to be thread-safe if it keeps state across requests. Maybe I thought this was more obvious than it actually is, so writing it down may help some users. May I open a new PR clarificating it or do you guys prefer to handle this yourselves?

kgadams:
Just go ahead, it's your baby :-)

PR #301, opened 2017-10-08
Fix WriteHeader in TestA301ResponseWriter.

leitzler:
WriteHeader did only set status field for a local copy that was discarded
upon return.


kisielk:
Thanks :)

PR #302, opened 2017-10-08
Clear up confusion around StrictSlash

### LINKED ISSUE #145: StrictSlash is confusing
This

"When true, if the route path is "/path/", accessing "/path" will redirect to the former and vice versa. In other words, your application will always see the path as specified in the route."

and

"When false, if the route path is "/path", accessing "/path/" will not match this route and vice versa."

should be inverted, because strict==true should mean no trailing slash allowed.
It's name and docs are confusing.

omar-h:
Fixes #145.

I've kept the StrictSlash function for backwards compatibility. The StrictRouting function is only there to serve as a clearer way to impose a strict slash.
It does the opposite of StrictSlash.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -163,6 +163,19 @@ func (r *Router) StrictSlash(value bool) *Router {
 	return r
 }
 
+// StrictRouting serves as a clearer version of StrictSlash. It does the
+// the opposite of StrictSlash.
+//
+// When true, if the route path is "/path", accessing "/path/" will not much
+// this route and vice verse.
+//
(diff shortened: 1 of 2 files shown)


elithrar:
I don’t think this adds enough value to afford adding to the API surface.

A PR to improve the docs on StrictSlash would be better.

omar-h:
Sure, you can just close this pr if you don't feel like it's good enough.

2017-10-09 6:02 GMT+02:00 Matt Silverlock <notifications@github.com>:

Issue #287, opened 2017-08-18
Ability to get the query template from a route


lukeng:
Hi,
I have a route with query inside, and I would like to generate a URL for reference but I can't find a way to retrieve the query template:
```
r := mux.NewRouter()
p := r.Path("/api").Queries("foo", "{foo}", "bar", "{bar}")
fmt.Println(p.GetPathTemplate()) // only returns "/api", but not "/api?foo={foo}&bar={bar}"
```

may it be better if GetPathTemplate() also returns the query template, or add a new GetQueriesTemplate() ?

paulbes:
We are also interested in this type of functionality, is it best to open a PR for this?

elithrar:
@paulbes I'd happily accept a PR for `GetQueriesTemplate()` functionality.

paulbes:
@elithrar, as a side-note, is there any intention for adding optional query parameters, e.g., a more concise way of achieving this: https://stackoverflow.com/questions/45378566/gorilla-mux-optional-query-values

elithrar:
No current plans, but open to a proposal that doesn't complicate the API.

mux has a large API, and the more we add, the higher the barrier to
newcomers as well as the more we need to document, maintain, test.

PR #304, opened 2017-10-11
[ISSUE-287] GetQueriesTemplates and GetQueriesRegexp extraction

paulbes:
Developers can now extract the query templates and regexps from a router as lists of combined query pairs.

## Description
The implementation returns lists of combined pairs of registered `Queries`; as templates or regexps. It was deemed more appropriate to return lists, at least for regexps, rather than a joined string of all combined pairs, since each query pair is considered a separate matcher, e.g., if we have the following: `Queries("foo", "{v1}", "baz", "{v2}")`, when joined using regexps we would end up with: `^foo=(?P<v0>.*)$,^baz=(?P<v0>.*)$`. The multiple `v0` entries might be confusing. For the query template, we could consider joining, but even then it will be a bit strange, e.g., `foo={v1},baz={v2}`.

## Motivation and Context
Being able to extract all parts of a registered route, including queries, can be very useful. In particular when generating REST API documentation. In addition, it can be useful to determine that tests have been written for all registered routes; by matching recorded and registered routes with their various parts.

## How Has This Been Tested (if appropriate)? 
Added `testQueriesRegexp` and `testQueriesTemplates` methods and added `queriesTemplate` and `queriesRegexp` fields to `routeTest`. All `TestQueries` cases are using this functionality.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring (non-breaking changes, no bug fixing, no new features)
- [x] Testing (New unit/integration/performance tests)

## Checklist:

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.

## Related tickets or issues:
This resolves issue #287


elithrar:
Thanks for the detailed write-up: it helps a lot. The patch looks good to
me; @kisielk as a second pair of eyes before merging.

kisielk:
Is there a reason you'd want just the Regexp or just the templates? Is there value in it being two separate functions?

paulbes:
I can't speak for all use-cases, but we will end up using `GetQueriesTemplates` primarily, I added the `GetQueriesRegexp` for completeness primarily; there being an equivalent one for `Path`. I am not sure when it would be used, perhaps in use-cases where a developer wants to examine the final regexp that is being matched with by mux.

Are you considering merging them into one or dropping either? If merging, what would this look like?

paulbes:
Do we need to make any changes to this PR or should it be merged?

elithrar:
If @kisielk is OK with the API here then it’s OK to merge.

kisielk:
The symmetry with the `GetPathXXX` methods makes sense. Sorry I missed the last email update. Thanks for the patch!

PR #306, opened 2017-10-17
Use req.URL.EscapedPath() instead of getPath(req)

mikelikesbikes:
This change drops support of go < 1.5. go1.4 has been [officially unsupported](https://golang.org/doc/devel/release.html#policy) since go1.6 was released 2016/02/17.

The PR was motivated by a bug I discovered, where `req.RequestURI` was returning `""`, so my router was failing to route on the escaped path.

Very open to thoughts and/feedback on how to proceed. Thanks!

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"path"
 	"regexp"
-	"strings"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


elithrar:
I am OK with this, and this makes more sense than the other PR.

1.5 has been around for a long while.

kisielk:
Agreed

PR #307, opened 2017-10-17
use req.URL.RequestURI() instead of req.RequestURI

mikelikesbikes:
This patches an issue where the `req.RequestURI` field is blank.

Alternative to #306.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -412,12 +412,12 @@ func setCurrentRoute(r *http.Request, val interface{}) *http.Request {
 // getPath returns the escaped path if possible; doing what URL.EscapedPath()
 // which was added in go1.5 does
 func getPath(req *http.Request) string {
-	if req.RequestURI != "" {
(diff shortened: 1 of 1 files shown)


mikelikesbikes:
(sorry for the noise... I expected this to be ok on the older go versions)

elithrar:
The docs state that the URL is parsed into this field, so should we not
handle that case?

https://golang.org/pkg/net/http/#Request

> For server requests the URL is parsed from the URI supplied on the
Request-Line as stored in RequestURI.

Thus, if req.URL.RequestURI() works, when does it behave differently from
the regular struct field?

I’d like to understand the fix before approving.

mikelikesbikes:
@elithrar I think this might be something in my service's tests. I'm going to close it.

elithrar:
Noted. Feel free to post back here if you need any help!

PR #311, opened 2017-11-03
MatchErr is set to ErrNotFound if NotFoundHandler is used

roobre:
As discussed in #293, it should be useful if `RouteMatch.MatchErr` were set to an error if no match is found, even if some subrouters have a custom `NotFoundHandler`.

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -1877,6 +1877,96 @@ func TestSubrouterHeader(t *testing.T) {
 	}
 }
 
+func TestNoMatchMethodErrorHandler(t *testing.T) {
+	func1 := func(w http.ResponseWriter, r *http.Request) {}
+
+	r := NewRouter()
+	r.HandleFunc("/", func1).Methods("GET", "POST")
+
+	req, _ := http.NewRequest("PUT", "http://localhost/", nil)
+	match := new(RouteMatch)
+	matched := r.Match(req, match)
+
+	if matched {
+		t.Error("Should not have matched route for methods")
+	}
+
+	if match.MatchErr != ErrMethodMismatch {
+		t.Error("Should get ErrMethodMismatch error")
+	}
+
+	resp := NewRecorder()
+	r.ServeHTTP(resp, req)
+	if resp.Code != 405 {
+		t.Errorf("Expecting code %v", 405)
+	}
+
+	// Add matching route
+	r.HandleFunc("/", func1).Methods("PUT")
+
+	match = new(RouteMatch)
+	matched = r.Match(req, match)
+
+	if !matched {
+		t.Error("Should have matched route for methods")
+	}
+
+	if match.MatchErr != nil {
+		t.Error("Should not have any matching error. Found:", match.MatchErr)
+	}
+}
+
+func TestErrMatchNotFound(t *testing.T) {
+	emptyHandler := func(w http.ResponseWriter, r *http.Request) {}
+
+	r := NewRouter()
+	r.HandleFunc("/", emptyHandler)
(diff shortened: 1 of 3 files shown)


mux_test.go
```diff
@@ -1982,3 +1982,54 @@ func TestNoMatchMethodErrorHandler(t *testing.T) {
...
+	s := r.PathPrefix("/sub/").Subrouter()
+	s.HandleFunc("/", emptyHandler)
+
+	// Regular 404 not found
+	req, _ := http.NewRequest("GET", "/sub/whatever", nil)
+	match := new(RouteMatch)
+	matched := r.Match(req, match)
+
+	if matched {
+		t.Error("Subrouter should not have matched that")
```
elithrar: I would add some context - `”subrouter should not have matched: got %v”, route”` to aid debugging.

mux_test.go
```diff
@@ -1982,3 +1982,54 @@ func TestNoMatchMethodErrorHandler(t *testing.T) {
...
+	// Now lets add a 404 handler to subrouter
+	s.NotFoundHandler = http.NotFoundHandler()
+	req, _ = http.NewRequest("GET", "/sub/whatever", nil)
+
+	// Test the subrouter first
+	match = new(RouteMatch)
+	matched = s.Match(req, match)
+	// Now we should get a match
+	if !matched {
+		t.Error("Subrouter should have matched that")
```
elithrar: Here too.

mux_test.go
```diff
@@ -1982,3 +1982,54 @@ func TestNoMatchMethodErrorHandler(t *testing.T) {
...
+		t.Errorf("Expected ErrNotFound MatchErr, but was %v", match.MatchErr)
+	}
+
+	// Now test the parent (MatchErr should propagate)
+	match = new(RouteMatch)
+	matched = r.Match(req, match)
+
+	// Now we should get a match
+	if !matched {
+		t.Error("Subrouter should have matched that")
```
elithrar: Here too.

roobre:
I've added the matching route where none is expected but one is got. I've also moved the last two test cases above the private functions, which I think is more appropriate.

The go fmt issue with travis is bugging me a bit, but afaik it's not my fault. It appeared when I merged upstream :/

elithrar:
Thanks for this. Let me see why gofmt is failing [only] on tip.

elithrar:
Rebase against master to pull in https://github.com/gorilla/mux/pull/312

route.go
```diff
@@ -72,7 +72,11 @@ func (r *Route) Match(req *http.Request, match *RouteMatch) bool {
 		return false
 	}
 
-	match.MatchErr = nil
+	if match.MatchErr == ErrMethodMismatch {
+		// We found a non-mismatching route, clear MatchErr
```
kisielk: non-mismatching is a bit of a double negative :) how about just "matching" so it's less confusing?

kisielk:
LGTM other than comment

route.go
roobre: Hm, I used that wording to make clear that the ErrMethod*Mismatch* found on the previous iteration no longer applies. But I guess I can reword it a bit.

I'll do it tomorrow tho, its 2AM here and I'm quite asleep.

kisielk:
Thanks! I don't mean to be too nitpicky, but the code base is already pretty hard to follow so any little bit helps.

roobre:
@kisielk No prob! I'm not a native speaker, so I can make some grammar mistakes here and there. Also, I'd do the same on projects I mantain :P

PR #313, opened 2017-11-05
Document router.Match

kisielk:
The return values are getting confusing. Hopefully this helps.


elithrar:
Great idea, thanks @kisielk!

kisielk:
Really makes me realize that the passing match by pointer and bool return value is a pretty ugly API :(

elithrar:
Tagged with v2 (aka "hopes and dreams")

PR #314, opened 2017-11-09
docs) fix outdated UseEncodedPath method docs

chrisdostert:
https://github.com/gorilla/mux/pull/306 changed UseEncodedPath to use native go encoded path handling so cautions in it's docs are no longer applicable.


elithrar:
Thanks @chrisdostert!

Issue #300, opened 2017-10-02
Subrouters match the wrong method when more than one matches


silverweed:
Introduced by https://github.com/gorilla/mux/pull/288.
Given the following code:

```go
package main

import (
	"fmt"
	"net/http"

	"github.com/gorilla/mux"
)

func main() {
	router := mux.NewRouter()
	get := router.Methods("GET", "HEAD").Subrouter()
	post := router.Methods("POST").Subrouter()

	post.HandleFunc("/some/{thing}", func(rw http.ResponseWriter, req *http.Request) {
		fmt.Fprintln(rw, "post")
	})
	get.HandleFunc("/some/{thing}", func(rw http.ResponseWriter, req *http.Request) {
		fmt.Fprintln(rw, "get")
	})

	http.Handle("/", router)
	http.ListenAndServe(":8888", nil)
}
```

a POST request to /some/thing will always match the `get` subrouter:

```sh
curl localhost:8888/some/thing -X POST # -> get
```

The same happens when there is a subrouter with a "catch-all" route, like 
```go
get.HandleFunc("/{anything}", getAnything)
```
This route will match any valid POST route *without parameters*. 

These behaviours happen no matter the calling order of the `HandleFunc`s.

Commenting [these lines](https://github.com/gorilla/mux/blob/master/route.go#L61-L64) seems to solve the issue, but of course breaks #288

leofachini:
I'm facing the same problem as you @silverweed, but with another route setup. (Would you think these might be related? If not, I will open a new issue.)

With the following code:

`func main() {
	r := mux.NewRouter()
	sub := r.PathPrefix("/").Subrouter()

	sub.StrictSlash(true).Path("/test").Methods("GET").HandlerFunc(getFunc)
	sub.StrictSlash(true).Path("/test").Methods("POST").HandlerFunc(postFunc)

	log.Fatal(http.ListenAndServe(":3000", r))
}

func getFunc(w http.ResponseWriter, r *http.Request) {

	fmt.Printf("%+v\n", r.Method)

	json.NewEncoder(w).Encode(r.Method)
}

func postFunc(w http.ResponseWriter, r *http.Request) {

	fmt.Printf("%+v\n", r.Method)

	json.NewEncoder(w).Encode(r.Method)
}
`

A GET to localhost:3000/test will result in:
`GET` :heavy_check_mark:

A GET to localhost:3000/test/ will result in:
`GET` :heavy_check_mark:

A POST to localhost:3000/test will result in:
[...]

PR #317, opened 2017-11-27
[bugfix] Fix method subrouter handler matching (#300)

mtso:
Fix handler match for subrouters created from method-matcher-based routes where the paths match (issue #300).

- Add a test case for routers with methods-based subrouters and matching paths
- Pass the test case with an addition to `Route.Match`

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -1968,310 +1968,269 @@ func TestErrMatchNotFound(t *testing.T) {
 }
 
 // methodsSubrouterTest models the data necessary for testing handler
-// matching for subrouters created with set HTTP methods.
+// matching for subrouters created after HTTP methods matcher registration.
 type methodsSubrouterTest struct {
-	title      string
-	router     *Router
-	flags      []bool
-	methods    []string
-	method     string
-	path       string
+	title    string
+	wantCode int
+	router   *Router
+	// method is the input into the request and expected response
+	method string
+	// input request path
+	path string
+	// redirectTo is the expected location path for strict-slash matches
 	redirectTo string
-	wantCode   int
 }
 
 // TestMethodsSubrouter tests the correct matching of handlers for
 // subrouters registered after a route's method-matcher is set.
 func TestMethodsSubrouter(t *testing.T) {
-	flags1 := make([]bool, 3)
+	MethodGet := "GET"
+	MethodPost := "POST"
+	MethodPut := "PUT"
+	MethodPatch := "PATCH"
+	MethodDelete := "DELETE"
+
+	getHandler := func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(MethodGet)) }
+	postHandler := func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(MethodPost)) }
+	postHandler2 := func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(MethodPost + "2")) }
+	putHandler := func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(MethodPut)) }
+	patchHandler := func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(MethodPatch)) }
+	deleteHandler := func(w http.ResponseWriter, r *http.Request) { w.Write([]byte(MethodDelete)) }
+
 	router1 := NewRouter()
-	methods1 := []string{"GET", "POST", "DELETE"}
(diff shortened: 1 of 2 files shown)


mux_test.go
```diff
@@ -1967,6 +1967,314 @@ func TestErrMatchNotFound(t *testing.T) {
...
+	methods    []string
+	method     string
+	path       string
+	redirectTo string
+	wantCode   int
+}
+
+// TestMethodsSubrouter tests the correct matching of handlers for
+// subrouters registered after a route's method-matcher is set.
+func TestMethodsSubrouter(t *testing.T) {
```
elithrar: It may be better to break this up a little, to allow parallel test execution (via `t.Parallel()`),.

mux_test.go
```diff
@@ -1967,6 +1967,314 @@ func TestErrMatchNotFound(t *testing.T) {
...
+	flags1 := make([]bool, 3)
```
elithrar: It isn't obvious (e.g. at first) what the purpose of the slice of bools / flags are for at first. Can you add a comment here to aid future contributors, and/or (better) look at a better way to test these that doesn't require this state tracking?

mux_test.go
mtso: Ok, I can try that.

mux_test.go
mtso: Yes, I can try that, would inspecting the response body be better?

kisielk: I actually don't quite understand it, is it just to show which methods matched? Why not use a `map[string]bool` for this?

mtso: Yes, the flags are used to check which handler is invoked. I wanted to check sequentially (for when two same routes are registered, but only the first one should be matched), but I can use `map[string]bool` with `[method][number]` like `get1` and `get2`.

kisielk: I would also call it something like `matched_methods` instead of just `flags`. The tests for this package are pretty confusing and convoluted, so anything that helps add some clarity is a big plus.

mtso:
Broke up `TestMethodsSubrouter` into five separate test cases based on the router initialized for each one. `t.Parallel` is also invoked in each one, so they run concurrently. Breaking them up further without `t.Run` would make the router initialization strategy more verbose.

I also removed the `methodsSubrouterTest` statefulness so the handler-match assertions are cleaner.

mux_test.go
```diff
@@ -1967,6 +1967,330 @@ func TestErrMatchNotFound(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		testMethodsSubrouter(t, test)
+	}
+}
+
+// TestMethodsSubrouterPathPrefix matches handlers on subrouters created
+// on a router with a path prefix matcher and method matcher.
```
elithrar: 👍

mux_test.go
```diff
@@ -1967,6 +1967,330 @@ func TestErrMatchNotFound(t *testing.T) {
...
+func testMethodsSubrouter(t *testing.T, test methodsSubrouterTest) {
+	// Execute request
+	req, _ := http.NewRequest(test.method, test.path, nil)
+	resp := NewRecorder()
+	test.router.ServeHTTP(resp, req)
+
+	switch test.wantCode {
+	case http.StatusMethodNotAllowed:
+		if resp.Code != http.StatusMethodNotAllowed {
+			t.Errorf("(%s) Expected \"405 Method Not Allowed\", but got %d code", test.title, resp.Code)
```
elithrar: Use backticks (`) for raw strings so you don't have to escape quotes.

mtso:
The relevant string literals have been updated to use backticks. Thanks for both your helpful feedback!

mux_test.go
```diff
@@ -1967,6 +1967,330 @@ func TestErrMatchNotFound(t *testing.T) {
...
+const (
+	methodGet    = "GET"
+	methodPost   = "POST"
+	methodPut    = "PUT"
+	methodPatch  = "PATCH"
+	methodDelete = "DELETE"
+)
+
+// Method handlers return the method string.
+var (
```
kisielk: I think this would make it more clear:

```
func methodHandler(method string) http.HandlerFunc {
        return method
   }
}
```

That way when you're defining the test fixtures, the string returned by the handler is right there in the fixture as opposed to these global functions.

mtso: Ok, makes sense. I noticed the HTTP methods in other tests are raw strings. Should these constants be replaced, too?

mtso:
The global method handler variables and constants have been removed.

kisielk:
LGTM! Thanks

elithrar:
Thanks for the contribution @mtso!

Issue #308, opened 2017-10-25
Unusual path parameter values make mux to choose incorrect handler


sergey-seleznev:
Hi!

I would like to define a number of handlers, that have: 
- same URL with a file path as the rightmost parameter,
- different HTTP request methods.

Here's a simplified example:
```
func main() {
	router := mux.NewRouter()
	router.HandleFunc("/files/{path:.*}", getHandler).Methods("GET")
	router.HandleFunc("/files/{path:.*}", postHandler).Methods("POST")
	log.Fatal(http.ListenAndServe(":8080", router))
}

func getHandler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("[GET] " + mux.Vars(r)["path"]))
}

func postHandler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("[POST] " + mux.Vars(r)["path"]))
}
```

It seems that `mux` tends to ignore methods and select improper handler in some specific cases:
- GET `http://localhost:8080/files/file.txt` => `[GET] file.txt` (OK)
- POST `http://localhost:8080/files/file.txt` => `[POST] file.txt` (OK)
- GET `http://localhost:8080/files/path/to/file.txt` => `[GET] path/to/file.txt` (OK)
- POST `http://localhost:8080/files/path/to/file.txt` => `[POST] path/to/file.txt` (OK)
- GET `http://localhost:8080/files/path/to//file.txt` => `[GET] path/to/file.txt` (OK)
- POST `http://localhost:8080/files/path/to//file.txt` => `[GET] path/to/file.txt` (unexpected)
- GET `http://localhost:8080/files/path/to/./file.txt` => `[GET] path/to/file.txt` (OK)
- POST `http://localhost:8080/files/path/to/./file.txt` => `[GET] path/to/file.txt` (unexpected)
- GET `http://localhost:8080/files/path/to/../file.txt` => `[GET] path/file.txt` (OK)
- POST `http://localhost:8080/files/path/to/../file.txt` => `[GET] path/file.txt` (unexpected)

Handler declaration order doesn't change the tests results mentioned.

elithrar:
This may be related to https://github.com/gorilla/mux/issues/300

kisielk:
Can you check if the latest version fixes the problem for you?

sergey-seleznev:
[...]

PR #321, opened 2017-12-02
[docs] Note StrictSlash re-direct behaviour #308

elithrar:
* StrictSlash enabled routes return a 301 to the client
* As per the HTTP standards, non-idempotent methods, such as POST or PUT, will be followed with a GET by the client
* Users should use middleware if they wish to change this behaviour to return a HTTP 308.

Addresses https://github.com/gorilla/mux/issues/308

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -164,7 +164,7 @@ func (r *Router) GetRoute(name string) *Route {
 // StrictSlash defines the trailing slash behavior for new routes. The initial
 // value is false.
 //
-// When true, if the route path is "/path/", accessing "/path" will perform a  redirect
+// When true, if the route path is "/path/", accessing "/path" will perform a redirect
 // to the former and vice versa. In other words, your application will always
 // see the path as specified in the route.
 //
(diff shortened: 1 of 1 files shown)


mux.go
```diff
@@ -164,13 +164,18 @@ func (r *Router) GetRoute(name string) *Route {
 // StrictSlash defines the trailing slash behavior for new routes. The initial
 // value is false.
 //
-// When true, if the route path is "/path/", accessing "/path" will redirect
+// When true, if the route path is "/path/", accessing "/path" will perform a  redirect
```
elithrar: Need to fix. Can't do that in GitHub.

kisielk: ??

kisielk:
> Addresses #308

Interesting numerical coincidence :)

mux.go
elithrar: Introduced a space and can’t edit my branch via the web UI (not at my laptop) 😉

elithrar:
🙌🏼

Issue #233, opened 2017-02-25
Public test API to set URL params


achiku:
Thank you for creating such a wonderful library!! I have gone through issues/prs, and now put a proposal up. Any comments are welcomed!

### Proposal
How do you think about adding a function to set URL params in context, which is implicitly declared as an API only for test. Testing as a public API is borrowed from this slide [Advanced Testing with Go](https://speakerdeck.com/mitchellh/advanced-testing-with-go?slide=53).

### Background
First of all, as it's discussed in several issues, it's not good idea to make context key public or basic type since it may conflict with other keys set from different packages, and accidentally overwrite values. Secondally, it is also not good idea to make a `setVars` function public, since it is fundamentally not necessary for this library to work as intended, and widen its API boundary, which tend to increase difficulty to maintain a clean project.

However, it is still useful if library users can add URL params to context for testing. The main reason is that users can separate middlewares, and other routing mechanism aside when they test a function that implements `http.Handler` interface.

### Implementation

Below is just an example.

- https://github.com/achiku/mux/commit/f8f4828232971c798713d8c4f7bb9739f9950dc4

### Related issues/prs
* https://github.com/gorilla/mux/issues/187
* https://github.com/gorilla/mux/issues/167
* https://github.com/gorilla/mux/pull/112

elithrar:
What's the difference between a public SetVars() and this proposal? :)

achiku:
No difference in implementation, but clearly stating "this function is for test" by making it public test API like `TestSetURLParam(r *http.Request, val map[string]string) *http.Request`, the library can provide a feature making unit test finer grain without encouraging misuse of `context.Context` modification in production code.

andrefsp:
Is there any development on this? IMO this is a quite important issue to be solved as it makes unittest a lot more complicated than it should.

grim-fendango:
I think the provided implementation strikes a good balance between testibility and discouraging bad behaviour. Could we please get this into a PR and merged ASAP?

grim-fendango:
Hey peeps, I've cherry picked the commit provided by achiku and spun up a [PR](https://github.com/gorilla/mux/pull/322) would love to hear your feedback 😸

grim-fendango:
Hey we can close this now 🎉 since this [PR](https://github.com/gorilla/mux/pull/322) has been merged.

kisielk:
Right on :)

PR #322, opened 2017-12-07
Public test API to set URL params

grim-fendango:
As discussed in #233, it will be useful to have a test focused API to set URL params. Currently this is not possible because the route key type used to store the vars in the context is private and the setVars function is also private.

Currently there are two work arounds to address this, the first being to make pull the vars extraction up in to the middleware layer, which requires a significant refactor and the other is to spin up a server and running the request through a route to populate the context with the vars. The problem with this is that it's very heavy handed and for those running tests in hosted CI environments like travis, the extra bloat can slow down test execution potentially to the point where testing handlers no longer becomes viable.

This should allow for the testing of routes and route handlers to be done separately while also preserving the current level of indirection that has allowed for the underlying context implementation to be changed (as seen in this [PR](https://github.com/gorilla/mux/pull/169)).

The added public API end point is prefixed with "Test" and can be found in `testing.go` to emphasise that this is a utility specifically meant for testing rather than in normal program flow. This notion of testing as a public API is borrowed from this slide [Advanced Testing with Go](https://speakerdeck.com/mitchellh/advanced-testing-with-go?slide=53).

### DIFF
--- a/testing.go
+++ b/testing.go
@@ -1,0 +1,8 @@
+package mux
+
+import "net/http"
+
+// TestSetURLParam set url params
+func TestSetURLParam(r *http.Request, val map[string]string) *http.Request {
+	return setVars(r, val)
+}
--- a/testing.go
+++ b/testing.go
@@ -1,8 +1,0 @@
-package mux
-
-import "net/http"
-
-// TestSetURLParam set url params
-func TestSetURLParam(r *http.Request, val map[string]string) *http.Request {
-	return setVars(r, val)
-}
--- a/test_helpers.go
+++ b/test_helpers.go
@@ -6,11 +6,13 @@ package mux
 
 import "net/http"
 
-// SetURLVars sets URL variables. This can be used to simplify the testing of
-// request handlers.
-// Alternatively, URL variables can be set by making a route that captures the
-// required variables, starting a server and sending the request to that
-// server.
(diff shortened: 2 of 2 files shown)


testing.go
```diff
@@ -0,0 +1,8 @@
+package mux
+
+import "net/http"
+
+// TestSetURLParam set url params
```
elithrar: Could you expand on this? Why it exists, what the normal way of setting params is; URL should also be capitalized.

grim-fendango: Wow thanks for a fast response! 😃  I was just thinking about that myself. I'm just looking up some concrete examples for this sort of thing and then I'll reword it.

elithrar: Suggestion:

```
// TestSetURLParam sets the URL params for the given request, to be accessed via mux.Vars for testing route behaviour.
//
// This API should only be used for testing purposes; it provides a way to inject variables into the request context.
```

kisielk:
I think a better name is needed too. Starting it with `Test` implies that the function is a test, to me...

testing.go
grim-fendango: I just wrote something but yours is nicer. Thanks again.

grim-fendango:
@kisielk and @elithrar thanks again for your feedback 😺 . I think I've addressed you're concerns. Let me know if you find anything else.

PR #328, opened 2018-01-05
refactor routeRegexp, particularily newRouteRegexp.

kisielk:
The existing options matchPrefix, matchHost, and matchQueries are
mutually exclusive so there's no point in having a separate boolean
argument for each one. It's clearer if there's a single type variable.

strictSlash and useEncodedPath were also being passed as naked bools so
I've wrapped these in a struct called routeRegexpOptions for more clarity
at the call site.

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -14,6 +14,20 @@ import (
 	"strings"
 )
 
+type routeRegexpOptions struct {
+	strictSlash    bool
+	useEncodedPath bool
+}
+
+type regexpType int
+
+const (
+	regexpTypePath   regexpType = 0
+	regexpTypeHost   regexpType = 1
+	regexpTypePrefix regexpType = 2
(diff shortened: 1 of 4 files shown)


regexp.go
```diff
@@ -14,6 +14,20 @@ import (
...
 )
 
+type routeRegexpOptions struct {
+	strictSlash    bool
+	useEncodedPath bool
+}
+
+type regexpType int
+
+const (
```
elithrar: 👌🏼

elithrar:
Thanks for this Kamil! Would be nice for the both of us to continue to
refactor some of the internals: mux is a bit of a beast 😉

kisielk:
Definitely. I actually tried to refactor this code at least 4 times over the past few years, using interfaces and different types for matchers, but it never really worked out. This seems to at least make it a bit clearer for now.

PR #330, opened 2018-01-16
Make shutdown docs compilable

kushmansingh:
Just making the graceful shutdown example copy paste-able


kisielk:
Travis seems backlogged or stalled, but since this is only to the README I'm merging anyway..

Issue #334, opened 2018-01-20
SetURLVars accepts *http.Handler and returns *http.Handler


fharding1:
**What version of Go are you running?** (Paste the output of `go version`)
`go version go1.9.2 linux/amd64`

**What version of gorilla/mux are you at?** (Paste the output of `git rev-parse HEAD` inside `$GOPATH/src/github.com/gorilla/mux`)
`5ab525f4fb1678e197ae59401e9050fa0b6cb5fd`

**Describe your problem** (and what you have tried so far)
SetURLVars returns an *http.Handler and accepts an *http.Handler. It is idiomatic if you are going to accept a pointer, you should modify, and if you really do want to pass by value instead of reference, you should use http.Handler, not *http.Handler.

**Paste a minimal, runnable, reproduction of your issue below** (use backticks to format it)
```go
package main

import (
	"fmt"
	"net/http"
	"net/http/httptest"

	"github.com/gorilla/mux"
)

func main() {
	req, _ := http.NewRequest("GET", "/bla", nil)
	rr := httptest.NewRecorder()

	printVars := func(w http.ResponseWriter, r *http.Request) {
		fmt.Println(mux.Vars(r))
	}

	mux.SetURLVars(req, map[string]string{"test": "hello"})
	printVars(rr, req)

	// mux.SetURLVars(*req, map[string]string{"test": "hello"}) does not compile
	// printVars(rr, req)

	req = mux.SetURLVars(req, map[string]string{"test": "hello"})
	printVars(rr, req)

}
```

I would be more than happy to create a PR correcting this.

garyburd:
The function mux.SetURLVars returns a shallow copy of the argument.

fharding1:
Thanks for the quick response! In this circumstance if you are returning a modified shallow copy, what are the advantages of not just passing by value and modifying that? And even if you weren't, and actually modifying the http.Handler by pointer, what would be the point of returning the pointer again?

elithrar:
To be clear, it accepts (and returns) a `*http.Request`, not a `*http.Handler` (which would be a pointer to an interface): rarely useful.

Returning the pointer is a common pattern to allow method chaining. You can ignore the returned variable, but you could also call `mux.SetURLVars(r, vars).AddCookie(cookie)`.

garyburd:
The function argument and return type are *http.Request, not *http.Handler.

[...]

PR #335, opened 2018-01-20
[docs] Clarify SetURLVars

fharding1:
Clarify in documentation that SetURLVars does not modify the given *htttp.Request, provide an example of usage.

Addresses #334

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -2248,6 +2248,15 @@ func TestMethodsSubrouterPathVariable(t *testing.T) {
 	}
 }
 
+func ExampleSetURLVars() {
+	req, _ := http.NewRequest("GET", "/foo", nil)
+	req = SetURLVars(req, map[string]string{"foo": "bar"})
+
+	fmt.Println(Vars(req)["foo"])
+
+	// Output: bar
+}
+
 // testMethodsSubrouter runs an individual methodsSubrouterTest.
 func testMethodsSubrouter(t *testing.T, test methodsSubrouterTest) {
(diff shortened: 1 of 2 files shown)


kisielk:
Can you make the example a runnable example test? The function doc can be shorter, just "... does not modify its arguments, but returns a shallow copy..", it should be obvious that it can be assigned to something.

fharding1:
Sure! Just to clarify, would you want me to include package, imports, etc, or just what would be in the main? Like

```go
package main

import (
	"fmt"
	"net/http"

	"github.com/gorilla/mux"
)

func main() {
	req, _ := http.NewRequest("GET", "/foo", nil)
	req = mux.SetURLVars(req, map[string]string{"foo": "bar"})

	func(w http.ResponseWriter, r *http.Request) {
		fmt.Println(mux.Vars(r))
	}(nil, req)
}
```
vs
```go
req, _ := http.NewRequest("GET", "/foo", nil)
req = mux.SetURLVars(req, map[string]string{"foo": "bar"})

func(w http.ResponseWriter, r *http.Request) {
	fmt.Println(mux.Vars(r))
}(nil, req)

```

kisielk:
Make a function in one of the test files called `ExampleSetURLVars`  and put the whole example in there. That will show up in the documentation alongside the `SetURLVars` function. More details in the "Examples" section here: https://godoc.org/testing

fharding1:
Added. It's my first time using Example test functions, so I apologize if I messed up somewhere.

PR #336, opened 2018-01-20
README.md: add miss "time"

safeoy:
README.md file , Graceful Shutdown demo code miss "time" package import

https://github.com/gorilla/mux/blob/master/README.md#graceful-shutdown


kisielk:
Good catch, thanks!

PR #338, opened 2018-01-20
Document route.Get* methods consistently

kisielk:
They actually return an error instead of an empty list. `GetMethods` happened to not return an error, but it should for consistency, so I added that as well.


elithrar:
Thanks @kisielk!

elithrar:
Note: potentially minor breaking change for some users.

PR #341, opened 2018-01-22
Router.ServeHTTP(): fixed redirect issues if url.host/scheme/user are non-empty

mreithub:
I'm using your `handlers.CombinedLoggingHandler` for logging.  
To log the request user, I set `req.URL.User = url.User(...)` in a wrapper before invoking the logging handler.

This works fine for logging, but breaks `mux.StrictSlash(true)`.

Instead of redirecting to `/target/path`, it redirects to `user@/target/path`.

This PR fixes that by setting the Host, Schema and User fields (=> everything before the Path) of the `req.URL` copy to their `nil` values.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -129,6 +129,14 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			// http://code.google.com/p/go/issues/detail?id=5252
 			url := *req.URL
 			url.Path = p
+
+			// clean up URL (in case it's been modified for some reason)
+			// If there was a .User for example, url.String() would return user@/path,
+			// breaking the redirect
+			url.Schema = ""
+			url.Host = ""
+			url.User = nil
+
 			p = url.String()
 
 			w.Header().Set("Location", p)
--- a/mux.go
+++ b/mux.go
@@ -130,12 +130,13 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			url := *req.URL
 			url.Path = p
 
-			// clean up URL (in case it's been modified for some reason)
-			// If there was a .User for example, url.String() would return user@/path,
-			// breaking the redirect
-			url.Schema = ""
-			url.Host = ""
-			url.User = nil
+			if url.Host == "" {
+				// clean up non-absolute URLs (in case it's been modified, e.g. to have the username in the logs - see PR #341)
+				// If there was a .User for example, url.String() would return `//user@/path/to/file`,
+				// breaking the HTTP location header
+				url.Scheme = ""
+				url.User = nil
+			}
 
 			p = url.String()
 


kisielk:
Won't this break if using eg: HTTPS since schema is being reset?

mreithub:
The Scheme, Host and User fields are usually empty in `http.Request.URL`.

`req.URL.String()` will therefore return something like: `/path/to/file.html?q=123`

which will in turn result in a response like:

    HTTP/1.1 301 Moved Permanently
    Location: /path/to/file.html?q=123

(i.e. no scheme, user or hostname)
something every browser I know (even IE6 back in the day) is fine with.
They'll simply keep these values the same -> HTTPS stays HTTPS, etc...

To be fair, manually overriding the request user isn't documented in `handlers.CombinedLoggingHandler`, but its source does evaluate the value - and there's no other way I can think of to get it to log the username

mreithub:
According to the [MDN's Location header page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location):

    <url>: A relative (to the request URL) or absolute URL.


But if you suspect the patch might cause issues, I could add a check for `url.Host == ""`.

Here's a quick play.golang.org snippet with a few hand-crafted URLs: https://play.golang.org/p/lsGWk-0sfsl

Btw. I hope the travis build failing is just an issue of missing secrets (or something like that)

mreithub:
Ok - sorry for spamming - according to [this SO post](https://stackoverflow.com/a/6911635), [the golang doc for `http.Request.URI`](https://golang.org/src/net/http/request.go?s=2888:10965#L106) and [RFC 2616, section 5.1.2](https://www.ietf.org/rfc/rfc2616.txt),
`Scheme` and `Host` may not be empty (in case of proxy requests).

I've just added the `url.Host == ""` check, that should prevent any issues.

Oh, and I've fixed a typo in the last commit, sry about that.  
Now the tests pass again.

kisielk:
Looking at it now, maybe this whole section of code should just use `http.Redirect` instead, which deals with these issues already.

Seems the whole block could be replaced by just `http.Redirect(w, r, p, http.StatusMovedPermanently)`

mreithub:
Wouldn't that strip the query string? (since `p` is only the `URL.Path`)

The lines 

```
w.Header().Set("Location", p)
w.WriteHeader(http.StatusMovedPermanently)
```

should probably be replaced by the `http.Redirect()` call though - to fix encoding issues (unless that's been done somewhere earlier already)

kisielk:
The query string is in the original request which is passed to `http.Redirect`. The function takes care of preserving that, you can verify this in the source: https://golang.org/src/net/http/server.go?s=58879:58944#L1959

kisielk:
@elithrar any thoughts on just simplifying the code after `cleanPath` down to `http.Redirect` ?

mreithub:
I'm sorry, I just noticed that I've tried fixing my issue at the wrong place in the code (I assumed that was the part responsible for the strictSlash redirects).

Now I've found out that adding the following lines to `regexp.go:294` will fix my issue:

```go
// Prevent malformed redirect URLs (if for example some outer Handler set u.User)
if u.Host == "" {
  u.Scheme = ""
  u.User = nil
}
```

Should I close this PR?  
How are your thoughts on opening another one for the actual fix?

mreithub:
I had trouble actually testing the changes (other than `go test`) - and only later found out that was because I changed the code in `~/go/src/...` while my project was using `vendor/`.

Again, sorry for wasting your time on this.

kisielk:
I would prefer to go the `http.Redirect` route if it makes sense. There's already a lot of special cases in the code and I'd prefer to not add more.

mreithub:
I'll close this then.

Thanks a lot for your time (and quick responses)

PR #342, opened 2018-01-26
Added documentation for how to test path variables

palsivertsen:


palsivertsen:
I think we are on the same page regarding beginner friendliness :)

But I think I'm missing something. Please correct me if I'm wrong:
When the readme mentions `variables` it's referring to path variables. `mux.Vars` is how you retrieve variables from a request. There are two ways of setting a path variable, add a new route with `/path/{myVar}` path syntax or use `mux.SetURLVars`.

I don't see how the current example illustrates how to use path variables when it does not include any of `mux.Vars`, `/path/{myVar}` or `mux.SetURLVars`?

PR #345, opened 2018-02-17
Expose ordered route param variable names

npgm:
This PR exposes an ordered list of route param variable names. Existing APIs return insufficient information to determine the order of vars in a path: mux.Vars() returns an unordered map, GetPathTemplate() returns an unparsed path template string.

### DIFF
--- a/route.go
+++ b/route.go
@@ -618,6 +618,22 @@ func (r *Route) GetPathRegexp() (string, error) {
 	return r.regexp.path.regexp.String(), nil
 }
 
+// GetVarNames returns the route path variable names.
+// An error will be returned if the route does not define a path.
+func (r *Route) GetVarNames() ([]string, error) {
+	if r.err != nil {
+		return nil, r.err
+	}
+	if r.regexp == nil || r.regexp.path == nil {
+		return nil, errors.New("mux: route does not have a path")
+	}
+	var varsN []string
+	for _, varN := range r.regexp.path.varsN {
+		varsN = append(varsN, varN)
+	}
+	return varsN, nil
+}
+
 // GetQueriesRegexp returns the expanded regular expressions used to match the
 // route queries.
 // This is useful for building simple REST API documentation and for instrumentation


elithrar:
Whilst I’m not opposed to this, can you explain the use-case more clearly?
Does the name not provide sufficient context here?

npgm:
A route path in general is an ordered sequence, and a name alone does not indicate the position in that sequence. 

A concrete example: mapping route params to function arguments.

elithrar:
Can you show me a concrete example here, where unique names (which also
represent function arguments) do not suffice?

For added context: mux has a large and useful API. Adding to that API
increases the learning curve & maintenance burden, and thus we can't always
accept every feature request at face value. It seems that there's another
way to achieve what you're after.

npgm:
Golang's reflect package lets you inspect function argument types and position in the parameter list, but doesn't let you inspect the name of the argument. So if you want to map route parameters to function parameters, the only way to achieve this mapping (from what I can gather) is to rely on relative position. 

See `NumIn() int` and `In(i int) Type`: https://golang.org/pkg/reflect/#Type

However beyond this specific use case, it makes sense to me that this would be exposed in the API because a path is an ordered sequence. 

```
func Bind(fn interface{}) http.HandlerFunc {
	var (
		fninputs []reflect.Type
	)

	fnval := reflect.ValueOf(fn)
	fntyp := fnval.Type()

         // gather the types of the function's inputs 
	for i := 0; i < fntyp.NumIn(); i++ {
		argtyp := fntyp.In(i)
		fninputs = append(fninputs, fntyp.In(i))
	}

		route := mux.CurrentRoute(r)
		vars := mux.Vars(r)
		varnames, _ := route.GetVarNames()
		if len(varnames) !=  len(fninputs) {
			http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
		}

		curVarName := 0
		fnargs := make([]reflect.Value, len(fninputs))
 
                 // for each function input's type, cast the corresponding var to the type and add it to the arg slice
		for i, vtyp := range fninputs {
			switch i {
                          / * snip some irrelevant logic from this switch */
			default:
				vn := varnames[curVarName]
				curVarName += 1

                                  // this converts the string to the type (not included here)
				vcast, err := cast(vtyp, vars[vn]) 

				if err != nil {
					http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
				}

				fnargs[i] = reflect.ValueOf(vcast).Convert(vtyp)
			}
		}

                 // Finally, call the function with the args
		_ = fnval.Call(fnargs)
	}
}
```

npgm:
@elithrar is there any other information that would be helpful? Do you think this is a valid use case?

elithrar:
Hi @npgm - I don't think this makes sense to add to the API.

* The name of the route variable provides sufficient context in nearly all cases: in your "function argument" example, names are useful. If you truly need ordering, you can use a naming scheme that covers this.
* Supporting the (very) narrow use-case of building routes via reflection is not worth the addition to the API
* [`*Route.GetPathTemplate`](https://godoc.org/github.com/gorilla/mux#Route.GetPathTemplate) can approximate this

PR #355, opened 2018-03-14
Make Use() variadic

jsvensson:
Enables neater syntax when chaining several middleware functions:

``` go
router.Use(mwFunc1, mwFunc2, mwfunc3)
```

Fully backwards-compatible when using a single function as argument.


kisielk:
Good idea, thanks.

PR #361, opened 2018-03-23
Add two methods to easily get body values to map

okabbas:
Hello . In normal mode, we can get variables using PostFormValue, but with these two methods, the values are injected internally into the map, and we can use the map outside of the http access.

It is also very painful to the database and it's much better than PostFormValue.

These two methods are very practical and make it a lot of work for me and prevent long lines of extra code.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -5,6 +5,7 @@
 package mux
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


elithrar:
As mux is focused on routing requests, these methods don’t make much sense, and are both fairly small. I should also note that both functions don’t return the error and instead swallow it.

https://github.com/gorilla/schema can do this out of the box, if needed.

Issue #353, opened 2018-03-12
Add Access-Control-Allowed-Methods with mux.Route.Methods()


fharding1:
**What version of Go are you running?** (Paste the output of `go version`)
go version go1.9.3 linux/amd64

**What version of gorilla/mux are you at?** (Paste the output of `git rev-parse HEAD` inside `$GOPATH/src/github.com/gorilla/mux`)
c0091a029979286890368b4c7b301261e448e242

**Describe your problem** (and what you have tried so far)
When using mux.Route.Methods() to set allowed methods, there is no easy way to add CORS. I had a middleware for this, but you run into problems when you have two of "the same" URLs, for example `/reports/{eventKey}/{matchKey}` (PUT) and `/reports/{eventKey}/{team}` (GET). It would be nice if there was a simple way to integrate CORS with .Methods(), or if there is one already if it were more clear in the documentation.

```go
r.Handle("/reports/{eventKey}/{matchKey}", submitReport).Methods("PUT")
r.Handle("/reports/{eventKey}/{team}", getReportsByEventAndTeam).Methods("GET")
```

An OPTIONS request or request to either endpoint should have Access-Control-Allow-Methods: "GET", "PUT", "OPTIONS".

**Paste a minimal, runnable, reproduction of your issue below** (use backticks to format it)
N/A

elithrar:
You could implement this as middleware by calling `route.GetMethods()` and then setting the response header. Does that not achieve your need to dynamically set the header?

e.g.

```go
func AutoCORS(next http.Handler) http.Handler {
	fn := func(w http.ResponseWriter, r *http.Request) {
		route := mux.CurrentRoute(r)
		methods, err := route.GetMethods()
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}

		w.Header().Set("Access-Control-Allow-Methods", strings.Join(methods,
			","))
		next.ServeHTTP(w, r)
	}

	return http.HandlerFunc(fn)
}
```

fharding1:
@elithrar Thank you for the response! That somewhat works I think, but OPTIONS requests still don't work right. Because the two URL's are the same format, an OPTIONS request to /reports/{eventKey}/{matchKey} or /reports/{eventKey}/{team} returns **either** PUT, OPTIONS or GET, OPTIONS (assuming both .Methods() include OPTIONS) when they should return GET, PUT, OPTIONS. Is there any way around this? To accumulate all methods on "the same" format URL's?

[...]

PR #366, opened 2018-04-23
Add MethodMiddleware

fharding1:
Implements the behavior discussed in #353 with @elithrar. I'm not sure if this is the best way to only match based on path only.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"path"
 	"regexp"
+	"strings"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


elithrar:
Thanks @fharding1 - can you add tests to this as well?

fharding1:
@elithrar :+1: done.

fharding1:
bump @elithrar

elithrar:
Thanks for the reminder. On my list for this week!

kisielk:
The usage of this seems a bit confusing to me. The idea is that the router will use itself as middleware?

fharding1:
Yeah, because you might only want to apply the middleware to certain routes. I'm not super familiar with mux beyond using it in personal projects with fairly normal use cases, would there be a better way to do this that you would recommend? @kisielk

kisielk:
Since it doesn't depend on any unexported fields of the router, I don't think it needs to be a method of the router itself. It could be a new middleware function or type that takes the router as an argument. That way we don't need to grow the already large API of the Router type.

fharding1:
Hmm, I implemented it as a method on Router as that's how @elithrar recommended it in #353: "I'd be open to a PR that can do this as a feature on Router.", but I also agree with what you've said. @elithrar ?

elithrar:
With the way it’s been implemented, it doesn’t need to be on the router at all.

I had assumed it would be similar to NotFoundHandler, where it is a field on the Router, or a toggle (Boolean).

fharding1:
Alright, I've removed it from `Router`, the usage now looks like `router.Use(MethodMiddleware(router))` which I think makes sense.

middleware.go
```diff
@@ -28,3 +31,43 @@ func (r *Router) Use(mwf ...MiddlewareFunc) {
 func (r *Router) useInterface(mw middleware) {
 	r.middlewares = append(r.middlewares, mw)
 }
+
+// MethodMiddleware returns a MiddlewareFunc intended for setting the CORS
+// Method header, Access-Control-Allow-Methods. It sets Access-Control-Allow-Methods
+// on a  request, by matching routes based only on paths. It also handles
+// OPTIONS requests, by settings Access-Control-Allow-Methods, and then
+// returning without calling the next http handler.
+func MethodMiddleware(r *Router) MiddlewareFunc {
```
elithrar: Two parts:

* I think the function name could be more descriptive here: `CORSMethodMiddleware` - something to indicate that it applies to a CORS-driven use-case.
* Do we want to build the routes on every request? (may not be avoidable to avoid stale data, but thinking out loud here)

mux_test.go
```diff
@@ -2315,6 +2315,14 @@ func stringMapEqual(m1, m2 map[string]string) bool {
 	return true
 }
 
+// stringHandler returns a handler func that writes a message 's' to the http
+// response writer.
```
elithrar: "to the http.ResponseWriter" or just "to the response".

middleware.go
fharding1: I agree with the first part. As for building the routes on every request, I think it would be fairly difficult to do, but I don't know the mux architecture super well either. I can't think of any clean solution that keeps our data fresh. It would be nice to not have to walk the router on each request though of course, so I'm open to any ideas!

fharding1:
Fixed! Thank you for maintaining gorilla/mux and helping me add this feature!

elithrar:
Thanks for the contribution. Apologies for the slow responses. Right in the middle of changing jobs, so have been extremely busy!

PR #370, opened 2018-05-01
Fix linter issues (docs)

elithrar:


middleware.go
```diff
@@ -12,7 +12,7 @@ type middleware interface {
 	Middleware(handler http.Handler) http.Handler
 }
 
-// MiddlewareFunc also implements the middleware interface.
+// Middleware also implements the middleware interface.
```
shurcooL: Maybe it'd be a good idea to remove the word "also":

```
// Middleware implements the middleware interface.
```

When reading documentation for this method, it might be confusing. "Middleware _also_ implements it? In addition to what?"

mux.go
```diff
@@ -13,8 +13,11 @@ import (
 )
 
 var (
+	// ErrMethodMismatch is returned when the error in the request does not match
```
shurcooL: I suspect "when the **error** in the request" should be "when the **method** in the request".

elithrar:
Fixes in https://github.com/gorilla/mux/pull/375

PR #382, opened 2018-05-25
Fix double OPTIONS allowed method in CORSMethodMiddleware

### LINKED ISSUE #381: CORS: OPTIONS call not handled
**What version of Go are you running?**
go version go1.10.2 darwin/amd64

**What version of gorilla/mux are you at?**
c85619274f5d55a67491010f883a91f568b3a9b0

**Describe your problem**

* Create a router
* Register a route with `.Methods("GET")`
* Make the router use the new CORSMiddleware
* Call `OPTIONS` on the route
* Expect a `200 OK` with the `Access-Control-Allowed-Methods` set to `GET,OPTIONS`; get instead a 405.

The code that is supposed to handle `OPTIONS` requests (below) is never reached, unless `OPTIONS` is a registered method for the route.

https://github.com/gorilla/mux/blob/c85619274f5d55a67491010f883a91f568b3a9b0/middleware.go#L64-L66

**Paste a minimal, runnable, reproduction of your issue below**

Add this test to `middleware_test.go`
```Go
func TestCORSMiddlewareOPTIONS(t *testing.T) {
	router := NewRouter()

	cases := []struct {
		path                   string
		response               string
		method                 string
		testURL                string
		expectedAllowedMethods string
	}{
		{"/g/{o}", "a", "POST", "/g/asdf", "POST,PUT,GET,OPTIONS"},
		{"/g/{o}", "b", "PUT", "/g/bla", "POST,PUT,GET,OPTIONS"},
		{"/g/{o}", "c", "GET", "/g/orilla", "POST,PUT,GET,OPTIONS"},
		{"/g", "d", "POST", "/g", "POST,OPTIONS"},
	}

	for _, tt := range cases {
		router.HandleFunc(tt.path, stringHandler(tt.response)).Methods(tt.method)
	}

	router.Use(CORSMethodMiddleware(router))

	for _, tt := range cases {
		rr := httptest.NewRecorder()
		req := newRequest("GET", tt.testURL)

		router.ServeHTTP(rr, req)

		allowedMethods := rr.HeaderMap.Get("Access-Control-Allow-Methods")

		if want, have := 200, rr.Result().StatusCode; have != want {
			t.Errorf("Expected status code %d, found %d", want, have)
		}

		if allowedMethods != tt.expectedAllowedMethods {
			t.Errorf("Expected Access-Control-Allow-Methods '%s', found '%s'", tt.expectedAllowedMethods, allowedMethods)
		}
	}
}
```

Note that this new test passes if `OPTIONS` is explicitly set as a method in the test-table:
```Go
		{"/g/{o}", "a", "POST", "/g/asdf", "POST,PUT,GET,OPTIONS"},
		{"/g/{o}", "b", "PUT", "/g/bla", "POST,PUT,GET,OPTIONS"},
		{"/g/{o}", "c", "GET", "/g/orilla", "POST,PUT,GET,OPTIONS"},
		{"/g", "d", "POST", "/g", "POST,OPTIONS"},
		{"/g", "", "OPTIONS", "/g", "POST,OPTIONS"},
		{"/g/{o}", "", "OPTIONS", "/g/orilla", "POST,PUT,GET,OPTIONS"},
```

If this is the desired behaviour, and the `OPTIONS` method has to be set while registering the route in order for the middleware to handle it, then it would be better to document it. But honestly, it doesn't sound desirable.

pierreprinetti:
and add a note on how Methods matchers have to be set in order for the
middleware to respod to OPTIONS calls.

Fixes #381

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -48,6 +48,10 @@ func CORSMethodMiddleware(r *Router) MiddlewareFunc {
 					if _, ok := m.(*routeRegexp); ok {
 						if m.Match(req, &RouteMatch{}) {
 							methods, err := route.GetMethods()
+							if err == ErrRouteHasNoMethods {
+								allMethods = []string{"OPTIONS", "GET", "HEAD", "POST", "PUT", "DELETE", "TRACE", "CONNECT"}
+								return nil
+							}
 							if err != nil {
 								return err
 							}
--- a/middleware.go
+++ b/middleware.go
@@ -33,9 +33,11 @@ func (r *Router) useInterface(mw middleware) {
 }
 
 // CORSMethodMiddleware sets the Access-Control-Allow-Methods response header
-// on a request, by matching routes based only on paths. It also handles
-// OPTIONS requests, by settings Access-Control-Allow-Methods, and then
-// returning without calling the next http handler.
+// on a request, by matching routes based only on paths.
+//
+// If a route in the router is set to handle OPTIONS calls, this middleware will
+// respond to OPTIONS calls with HTTP code "200 OK", the
+// Access-Control-Allow-Methods header and an empty body.
 func CORSMethodMiddleware(r *Router) MiddlewareFunc {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
--- a/middleware.go
+++ b/middleware.go
@@ -48,10 +48,6 @@ func CORSMethodMiddleware(r *Router) MiddlewareFunc {
 					if _, ok := m.(*routeRegexp); ok {
 						if m.Match(req, &RouteMatch{}) {
 							methods, err := route.GetMethods()
-							if err == ErrRouteHasNoMethods {
-								allMethods = []string{"OPTIONS", "GET", "HEAD", "POST", "PUT", "DELETE", "TRACE", "CONNECT"}
-								return nil
-							}
 							if err != nil {
 								return err
 							}
--- a/middleware.go
+++ b/middleware.go
@@ -61,7 +61,7 @@ func CORSMethodMiddleware(r *Router) MiddlewareFunc {
 			})
 
 			if err == nil {
-				w.Header().Set("Access-Control-Allow-Methods", strings.Join(append(allMethods, "OPTIONS"), ","))
+				w.Header().Set("Access-Control-Allow-Methods", strings.Join(allMethods, ","))
 
 				if req.Method == "OPTIONS" {
 					return
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -378,39 +378,27 @@ func TestCORSMethodMiddleware(t *testing.T) {
 	}
 }
 
+// A route without a Method filter will not trigger the middleware.
 func TestCORSMiddlewareOPTIONSWithoutMethodMatcher(t *testing.T) {
 	router := NewRouter()
 
-	cases := []struct {
-		path                   string
-		response               string
-		testURL                string
-		expectedAllowedMethods string
-	}{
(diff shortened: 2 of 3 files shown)


pierreprinetti:
If #378 was to be merged, these docs would have to be amended.

With the current logic, the OPTIONS preflight request will only be responded to by the CORSMethodMiddleware if the method is matched by a route. If instead the middlewares were to be triggered even without a match, the OPTIONS call would be responded anyway once CORSMethodMiddleware is `Use`d.

pierreprinetti:
See a preview of the changes [here](https://godoc.org/github.com/pierreprinetti/mux#CORSMethodMiddleware)

pierreprinetti:
Hmm wait there's a problem. I'm adding a test to get it

pierreprinetti:
OK, it seems like we have a couple problems.

### Without method matcher
If the methods are not set explicitly, the middleware should allow [a reasonable amount of methods](https://tools.ietf.org/html/rfc2616#section-5.1.1).

### With method matcher
In case of a route with explicit method matchers, "OPTIONS" should be specified. This means that the production code should not add it explicitly:
https://github.com/gorilla/mux/blob/c85619274f5d55a67491010f883a91f568b3a9b0/middleware.go#L62

Doing so would result in a duplicate.

pierreprinetti:
The tests now pass. ptal! :)

middleware.go
```diff
@@ -33,9 +33,11 @@ func (r *Router) useInterface(mw middleware) {
...
 
 // CORSMethodMiddleware sets the Access-Control-Allow-Methods response header
-// on a request, by matching routes based only on paths. It also handles
-// OPTIONS requests, by settings Access-Control-Allow-Methods, and then
-// returning without calling the next http handler.
+// on a request, by matching routes based only on paths.
+//
+// If a route in the router is set to handle OPTIONS calls, this middleware will
+// respond to OPTIONS calls with HTTP code "200 OK", the
+// Access-Control-Allow-Methods header and an empty body.
```
elithrar: The docs here should be clear that the middleware will respond with a default set of methods, and list those methods, unless a specific set of methods apply to that route.

However, the more I think on this, adding non-idempotent methods by default seems risky. Why not just safe + idempotent (yes, I'm aware that people break these semantics...) methods - `OPTIONS, GET, HEAD, TRACE` - as per https://tools.ietf.org/html/rfc7231#section-8.1.3 - ?

fharding1:
Don't mean to butt in, but why would we want to respond with a default set of methods? If the developer does not use .Methods() on a route, it doesn't make sense to set the Access-Control-Allow-Methods header on it to me. Not even just safe and idempotent ones.

fharding1:
I get your point, but it seems to me that if I didn't set Methods() on a route I am purposely not setting any CORS headers because setting CORS headers is tied to .Methods(). It seems like it would be better to force the developer to explicitly allow all methods for CORS. That being said, that's mostly my opinion and what you're saying is valid too.

pierreprinetti:
With [779aa8f](https://github.com/gorilla/mux/pull/382/commits/779aa8fa5fc458dd92ef64103bbb3f348d039535), a route without explicit Method filters will be ignored by the middleware. Please take a look!

pierreprinetti:
(I'm not sure the `documentation` label still applies)

pierreprinetti:
I'm closing this as #381 has been identified as a wontfix. Please consider reusing this code if the CORS middleware discussion comes up again.

In the meantime, I have implemented a custom solution in the codebase I was working on, as the API of this specific feature in `mux` is a bit too magic for me.

Thank you!

elithrar:
Thanks for the feedback Pierre. If we see a few more reports we can revisit
this conversation.

PR #383, opened 2018-05-28
fix #2, return full host:port info from getHost

### LINKED ISSUE #2: Host() with  a port fails to match when the host is supplied in the HTTP Host: header
Old issue: http://code.google.com/p/gorilla/issues/detail?id=51

According to section 14.23 of RFC 2616 the Host header can include the port number if the default value of 80 is not used.

See https://github.com/gorilla/mux/blob/master/mux_test.go#L119 for a commented-out test case that fails.

Potential fix is editing getHost() https://github.com/gorilla/mux/blob/master/regexp.go#L237 so it has a flag as to whether to strip off the port information. A new field would need to be probably need to be added somewhere such as routeRegexp.

The main reason I was using Host() match with a port in the first place is for trying to build reverse routes that include the absolute path as opposed to a relative one.

santsai:
I ran into #2, i think its safe to simply `return r.Host` because `r.URL.IsAbs()` case is already doing that and is covered by existing tests cases.

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -312,17 +312,13 @@ func (v routeRegexpGroup) setMatch(req *http.Request, m *RouteMatch, r *Route) {
 }
 
 // getHost tries its best to return the request host.
+// According to section 14.23 of RFC 2616 the Host header
+// can include the port number if the default value of 80 is not used.
 func getHost(r *http.Request) string {
 	if r.URL.IsAbs() {
 		return r.URL.Host
 	}
-	host := r.Host
-	// Slice off any port information.
-	if i := strings.Index(host, ":"); i != -1 {
-		host = host[:i]
-	}
-	return host
-
+	return r.Host
 }
 
 func extractVars(input string, matches []int, names []string, output map[string]string) {
--- a/mux_test.go
+++ b/mux_test.go
@@ -104,7 +104,15 @@ func TestHost(t *testing.T) {
 			path:        "",
 			shouldMatch: false,
 		},
-		// BUG {new(Route).Host("aaa.bbb.ccc:1234"), newRequestHost("GET", "/111/222/333", "aaa.bbb.ccc:1234"), map[string]string{}, "aaa.bbb.ccc:1234", "", true},
+		{
+			title:       "Host route with port, match with request header",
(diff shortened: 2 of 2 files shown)


regexp.go
```diff
@@ -312,17 +312,13 @@ func (v *routeRegexpGroup) setMatch(req *http.Request, m *RouteMatch, r *Route)
 }
 
 // getHost tries its best to return the request host.
+// According to section 14.23 of RFC 2616 the Host header
+// can include the port number if the default value of 80 is not used.
```
elithrar: Can you point me to the relevant section of the RFC that covers this requirement for the Host header? It is mentioned in the context of a URL, but seems more ambiguous in the context of a Host header. It is not immediately clear in 7230 either: https://tools.ietf.org/html/rfc7230#section-5.4

> 3.2.2 http URL
> The "http" scheme is used to locate network resources via the HTTP
   protocol. This section defines the scheme-specific syntax and
   semantics for http URLs.
>  http_URL = "http:" "//" host [ ":" port ] [ abs_path [ "?" query ]]
> If the port is empty or not given, port 80 is assumed. 

I'm cautious about making changes here as it may change the matching behaviour (read: breaking change) of existing users.

kisielk:
It's a bit of a scary change, but it makes sense in theory and I guess we can merge it if all the tests (plus additional ones) pass.

elithrar:
I'm OK to merge this. It will require:

* A minor version bump (tagged as such)
* A commit message that clarifies this - 

summary: getHost() now returns full host & port information. 
full commit message: Previously, getHost only returned the host. As it now returns the port as well, any `.Host` matches on a route will need to be updated to also support matching on the port for cases where the port is not 80 (scheme == "http") or 443 (scheme == "https").

cognusion:
Tracking down why my apps suddenly stopped doing host matching properly, and stumbled onto this. Seems like a big backward-breaking change for a minor release. At the very least, when the .Host() contains just a host, and not a port, then hostname:* should be assumed, I'd think.

elithrar:
I'm open to a PR for that.

cognusion:
PR #447 does the job efficiently.

harshavardhana:
This broke API compatibility with 1.6.x we have a regression in MinIO because of this.

Issue #17, opened 2013-04-16
Schemes() if not matching my test


rbucker:
// this code does not work

```
r := mux.NewRouter()
r.HandleFunc("/getdata", getdata_handler).
    Host("localhost").
    Methods("GET").
    Headers("Accept", "text/json").
    Schemes("http").
    Name("getdata")
```

// this code works

```
r := mux.NewRouter()
r.HandleFunc("/getdata", getdata_handler).
    Host("localhost").
    Methods("GET").
    Headers("Accept", "text/json").
    Name("getdata")
```

Here is the test:

```
        $.ajax({
            url: 'http://localhost:3000/getdata?callback=?',
            //headers: { "Accept": "text/json" },
            type: 'GET',
            dataType: 'json',
            jsonp: 'callback',
            success: function(data) {
                console.log(data);
                template = $('#tabletemplate').clone().html();
                console.log(template);
                snippet = Mustache.render(template, data);
                console.log(snippet);
                $('#workspace').empty().append(snippet);
            },
            error: function(  jqXHR,  textStatus,  errorThrown) { console.log('error', textStatus, errorThrown); },
            beforeSend: function(jqXHR){ jqXHR.setRequestHeader("Accept", "text/json"); }
        });
```

Issue #130, opened 2015-10-24
mux/v2: need feedback


moraes:
I'd like to start a conversation and brainstorm about the future of gorilla/mux.

Background: gorilla/mux has evolved to be the most featureful router in Go, but this came with a price: the API feels bloated and has unneeded complications. It is also far from shining in benchmarks compared to other routers (not that gorilla/mux is a bottleneck to care about, but not-so-shiny numbers give a bad impression to new users).

I'd like to take gorilla/mux to the next level with these goals in mind:

1. Slim API;
2. Intuitive usage: simple rules, minimized unexpected behavior;
3. Improved performance;

After playing around, I came up with an implementation draft that is a radical departure from what gorilla/mux is now, and to achieve the slim-intuitive-fast goals some cuts were made:

- Merged Path(), Handler() and HandlerFunc();
- Merged URL(), URLHost() and URLPath();
- Moved Scheme, Host and PathPrefix matching to the route declaration syntax;
- Dropped duplicated Router/Route methods everywhere;

...but the most controversial ones are probably:

- Dropped query, header and custom matching;
- Dropped support for regular expressions;

The initial result is a much slimmer mux which, besides the elemental features (scheme, host, path and method matching), offers some conveniences not easily found in other packages (subrouting, URL building), and a damn high speed (teaser: it almost beats net/http in static matching and knocks out everybody else in matching with variables; and, oh, it doesn't make a single memory allocation for any match — zero, zip, zilch, nada).

But this is not gorilla/mux. It is something else. Maybe it doesn't even exist! Well, maybe it does. Enough teasing for now: I'd like to hear your crazy ideas, frustrations and aspirations for gorilla/mux, and to ask how could we manage an eventual API breakage in the best possible way.

elithrar:
* I'm definitely for a slimmer API, even if it means (spitballing here!) a new package under the gorilla organization to allow mux to stay "stable" for existing users.
* Could add a note to the current mux (v1) README and a bunch of "this is how you would achieve X in mux v1 in the new mux v1" examples elsewhere to help push users to the new package.
* Coin a new name - `gorilla/multiplex` or even just `gorilla/router` (naming is hard)

Other thoughts:

* Merging Path, Handler and HandlerFunc is A++ in my books.
[...]

PR #384, opened 2018-05-28
fix scheme matching for path only requests

santsai:
Path only requests (with Host set in header) have empty `r.URL.Scheme`, and scheme matching will always fail. Test cases and fix added. related to #17

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -7,6 +7,7 @@ package mux
 import (
 	"bufio"
 	"bytes"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


santsai:
It seems not good to change existing `.Schemes()` as it will probably be a breaking change for existing users. Instead `Route.InferScheme()` is added to indicate a route should infer scheme.

santsai:
@elithrar Do you mean to update the `.Schemes()` comment or to do it only when some flag is enabled?

elithrar:
Update the schemes comment.

I am reticent to add a new method to the API with a similar name. It will
be confusing and the benefit seems minimal.

Given the risk of breaking existing users, it makes sense to hold off until
a potential v2.

kisielk:
I am also not sure if it's safe to change to this at this point, there's a lot of risk of breaking existing code.

elithrar:
I've marked this as v2 - which we can review as part of #130 at "some point" in the future.

PR #385, opened 2018-06-05
Update LICENSE file to reflect Google employee contributions.

elithrar:
As per https://opensource.google.com/docs/patching/#common-rules -

> If project authors are listed in a LICENSE, COPYING, AUTHORS or similar file, please add Google LLC. if it’s not already there. If Google Inc. is already listed then there is no need to change it to Google LLC. This step only applies if the project authors are listed somewhere.

Updating this to reflect a few boxes I need to tick.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2012 Rodrigo Moraes. All rights reserved.
+Copyright (c) 2012-2018 Rodrigo Moraes, Google LLC. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are


kisielk:
Hm, should we update this repo to say "The Gorilla Authors" and then use an AUTHORS file instead?

elithrar:
I'm happy to do that. I can update those PRs tonight to standardize:

- Update LICENSE files to say "The Gorilla Authors"
- Add an `AUTHORS` file to list contributors.

I'll do this for all of the major projects (mux, sessions, securecookie, csrf, context, handlers, schema) and then the long tail on an as-needed basis.

(I guess you can tell where I work now?)

kisielk:
Yeah it's something that's long overdue, there wasn't anything pushing for it before but I guess it's time to get a bit more formalized :)

Issue #326, opened 2018-01-02
Proposal: Drop Support for Go < 1.7 as of Go 1.10


elithrar:
**Summary**: Dropping Go < 1.7 will allow us to remove the shim for `gorilla/context`, simplifying the codebase and making ongoing maintenance easier.

* The aim of those proposal is to gather potential feedback and signal to users the intent for change
* Go 1.10 is scheduled for release in Feb 2018 - https://tip.golang.org/doc/go1.10
* I suspect a not-insignificant number of users are likely on Go 1.6 (Ubuntu 16.04 LTS default)

kisielk:
I wish there was some way to tell how many people are still compiling with old Go versions

roobre:
If Ubuntu LTS users are a concern, maybe we can wait to the release of 18.04, which is scheduled for April iirc.

elithrar:
I think that’s what we’ll likely do. Once dep is in the official toolchain
we can also break out a v2.0 🌟

elithrar:
I'm planning on doing this before the end of June.

elithrar:
See #391

PR #391, opened 2018-07-09
Drop support for Go < 1.7: remove gorilla/context

fharding1:
Remove gorilla/context for reasons outlined in issue #326.


elithrar:
You’ll need to:

- update `travis.yml` to remove versions < 1.7 
- update documentation to make it clear that clearing the context only applies when using > Go 1.7

fharding1:
Alright. Also, I'm not sure, but it seems like we can remove contextClear from context_native and the code block from mux.go 158:160. If we did that KeepContext would be basically useless (I think), but removing it would be breaking. Should I just put a comment on it that says this has no effect?

flibustenet:
Actually gorilla/context is required. 
I think it will be more explicit to update `go.mod` to reflect this now with a `require github.com/gorilla/context`, then release the last mux that require it and then remove it in this PR.

elithrar:
@fharding1 Did you want to make the adjustment here as per @flibustenet's comment?

fharding1:
I'm not sure what he means, why is it required?

elithrar:
It would provide Module support before the change.

I don’t think it’s explicitly required, but I hadn’t seem acknowledgement
either way.

fharding1:
I'm not super familiar with modules, but yeah, I don't think it's required. Wasn't planning on addressing that feedback unless @flibustenet can clarify why it's required.

flibustenet:
I was just thinking of the historical side. We could see it was required and not after. But the changelog is ok also, nothing more...

PR #400, opened 2018-09-03
Remove useless context code

KimMachineGun:


KimMachineGun:
Thank you for your fast answer.

But, I think that useless condition check and call empty body function are unnecessary.
I hope that they are removed on next release. 😄

Issue #409, opened 2018-10-12
Convert tests to use subtests with t.Run


kisielk:
Just a maintenance task, but it will make test reporting cleaner in case of failure:

https://blog.golang.org/subtests

kanozec:
I think I can take this, but I'm not familiar with the go test. I just have some practice before.
Could you give me some advice for which test file to refactor first.
I think that would be really helpful, thx.

kisielk:
Still valid

fharding1:
I'm happy to take this on for the remainder of the tests. I assume you tagged this as still valid because like https://github.com/gorilla/mux/blob/ed099d42384823742bba0bf9a72b53b55c9e2e38/middleware_test.go#L48 and some other places similar to that could use subtests? Also, should I update `old_test.go`, or is that unmaintained (per comment `// Old tests ported to Go1. This is a mess. Want to drop it one day.`).

fharding1:
Should this issue be closed? Seems like most every test that it makes sense for is using subtests now.

PR #415, opened 2018-10-30
Use subtests in mux_test.go

kanozec:
related to #409


kisielk:
That was a lot less work than I expected :) Thanks!

Issue #411, opened 2018-10-19
Generic subrouter is eating middlewares


s4t13:
**What version of Go are you running?** 
`go version go1.11 darwin/amd64`

**What version of gorilla/mux are you at?**
`e3702bed27f0d39777b0b37b664b6280e8ef8fbf`

**Describe your problem**

The "global" subrouter seems to eat all the middlewares. If I declare a subrouter with `PathPrefix("").Subrouter()`, the later executed handler don't pass through the middlewares.

If I declare the subrouter *after*, the execution is normal tho.

**Paste a minimal, runnable, reproduction of your issue below**

## Example 1

```package main

import (
	"log"
	"net"
	"net/http"

	"github.com/gorilla/mux"
)

func handle() http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		log.Println("Called handle")
	}
}

func middleware(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		log.Println("Called middleware")
		next.ServeHTTP(w, r)
	})
}

func main() {
	r := mux.NewRouter()

	// _ = r.PathPrefix("").Subrouter()

	r1 := r.PathPrefix("/1").Subrouter()
	r1.HandleFunc("", handle())
	r1.Use(middleware)

	server := &http.Server{Handler: r, Addr: net.JoinHostPort("", "3000")}
	log.Fatal(server.ListenAndServe())
}
```

In this setting, if I `GET /1` I get:
```
Called middleware
Called handle
```

## Example 2
I uncomment the line:
```
func main() {
	r := mux.NewRouter()

	_ = r.PathPrefix("").Subrouter()

	r1 := r.PathPrefix("/1").Subrouter()
	// ... same as previously
}
```

And I `GET /1` I get:
```
Called handle
```

## Example 3

Now if move the line after `r1` definition:
```
func main() {
	r := mux.NewRouter()

	r1 := r.PathPrefix("/1").Subrouter()
	r1.HandleFunc("", handle())
	r1.Use(middleware)

        _ = r.PathPrefix("").Subrouter()
	// ... same as previously
}
```
And I `GET /1` I get:
```
Called middleware
Called handle
```

s4t13:
It makes sense. But instead of having a `404`, `handler` is still called at the end of the line, without passing through `middleware`.

Sterd:
[...]

PR #417, opened 2018-10-31
Clear any match error or handler when iterating a router

lavoiesl:
Fixes https://github.com/gorilla/mux/issues/411

Added tests that showed the previous failing behaviour.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -83,6 +83,16 @@ type Router struct {
 // field of the match argument.
 func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 	for _, route := range r.routes {
+		// Since there is another matcher to try, clear any match error or handler.
+		// Otherwise, the state of one matcher can leak into the other.
+		//
+		// However, do not override ErrMethodMismatch.
+		// If it not set again by a route, that's the error we will bubble up.
+		if match.MatchErr != ErrMethodMismatch {
+			match.MatchErr = nil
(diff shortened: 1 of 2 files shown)


kisielk:
Doesn't this result in just eating any errors that come from earlier routes? What if the first route checked is the right path, but has a method mismatch, and the second route is the wrong path? The method mismatch logic won't be triggered in that case.

lavoiesl:
Very good point. It’s confusing because the `ErrMethodMismatch` logic is inverted.
I added an exception and a test for it.

It would probably be cleaner to move the allowed methods to a dedicated property, but this seems good enough for now.

elithrar:
We should move them to a dedicated property: it will make it easier to
maintain in the future.

lavoiesl:
As I said, I couldn't find a proper solution with a dedicated property that was working. It might take a more significant refactoring.

I had a very real bug related to this that took me a shameful amount of time to diagnose. I propose merging this as is, considering it works for the new test case without introducing a known regression, and revisit later.

kisielk:
I am more concerned about the unknown regressions, I don't believe the test suite is fully comprehensive. 

What happens if the MatchErr is `ErrNotFound` in the first case, but then that gets wiped out to try other routes, will the `NotFoundHandler` ultimately be called?

elithrar:
Closing in favor of #418

PR #418, opened 2018-11-03
Refactor custom handlers logic

### LINKED ISSUE #411: Generic subrouter is eating middlewares

lavoiesl:
Fixes #411
Closes https://github.com/gorilla/mux/pull/417

This is a more comprehensive attempt, streamlining some of the intricate logic.
It does introduce a BC-break since previously a non-match with a custom handler was considered a match, but I feel this is mostly an internal state that shouldn't affect end-users.

Plenty of tests were added.

---

custom handler:
- Do not report "matched" when finding a custom handler
- Bubble up the handler that was found and do not override the custom
  handler when applying default behavior.
- Now it "matches" if and only if MatchErr is nil

methodMatcher:
- Maintain only one methodMatcher
- Append methods to existing methodMatcher when Methods() is called again
- Set the ErrMethodMismatch in the methodMatcher

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,11 +491,11 @@ r.Use(amw.Middleware)
 
 Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to. Middlewares _should_ write to `ResponseWriter` if they _are_ going to terminate the request, and they _should not_ write to `ResponseWriter` if they _are not_ going to terminate it.
 
-#### Middlewares on route-not-found
+#### Middlewares and Routing
 
-Middlwares only run when a route is found. If the request can't be matched, including if the method is not allowed, the middlewares will be skipped.
+Middlewares only run when a route is found. If the request can't be matched, including if the method is not allowed, the middlewares will be skipped.
 
-This is also true if you specify a custom `MethodNotAllowedHandler` or `NotFoundHandler`.
+This is also true if you specify a custom `MethodNotAllowedHandler` or `NotFoundHandler`: middlewares will not be attached to these handlers.
 
 To _also_ run a middleware on a custom handler, apply it manually:
 
--- a/README.md
+++ b/README.md
@@ -491,6 +491,24 @@ r.Use(amw.Middleware)
 
 Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to. Middlewares _should_ write to `ResponseWriter` if they _are_ going to terminate the request, and they _should not_ write to `ResponseWriter` if they _are not_ going to terminate it.
 
+#### Middlewares on route-not-found
+
+Middlwares only run when a route is found. If the request can't be matched, including if the method is not allowed, the middlewares will be skipped.
+
+This is also true if you specify a custom `MethodNotAllowedHandler` or `NotFoundHandler`.
+
+To _also_ run a middleware on a custom handler, apply it manually:
+
+```go
+r := mux.NewRouter()
+h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+    io.WriteString(w, "Nothing to see here")
+    w.WriteHeader(http.StatusNotFound)
+})
+h = loggingMiddleware(h)
+r.NotFoundHandler = h
+```
+
 ### Testing Handlers
 
 Testing handlers in a Go web application is straightforward, and _mux_ doesn't complicate this any further. Given two files: `endpoints.go` and `endpoints_test.go`, here's how we'd test an application using _mux_.
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -212,6 +212,7 @@ func TestMiddlewareNotFound(t *testing.T) {
 func TestMiddlewareMethodMismatch(t *testing.T) {
 	mwStr := []byte("Middleware\n")
 	handlerStr := []byte("Logic\n")
+	mnaStr := []byte("Method not allowed")
 
 	router := NewRouter()
 	router.HandleFunc("/", func(w http.ResponseWriter, e *http.Request) {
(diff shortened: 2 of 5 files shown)


lavoiesl:
@kisielk @elithrar Would you like more information about this patch?

elithrar:
I'm travelling, so just haven't had time to review in proper given it
touches a bunch of internal structures. Will try to set some time aside on
my flight back.

lavoiesl:
No worries, ping me if you want more info on my thought process.

middleware_test.go
```diff
@@ -212,6 +212,7 @@ func TestMiddlewareNotFound(t *testing.T) {
 func TestMiddlewareMethodMismatch(t *testing.T) {
 	mwStr := []byte("Middleware\n")
 	handlerStr := []byte("Logic\n")
+	mnaStr := []byte("Method not allowed")
```
elithrar: Needs trailing newline.

mux_test.go
```diff
@@ -1950,13 +2021,21 @@ func TestNoMatchMethodErrorHandler(t *testing.T) {
 	}
 }
 
+// Below, we compare the handlers.
+// We need an actual struct pointer since Go doesn't support function comparison
+type emptyHandler struct {
```
elithrar: `type emptyHandler struct{}` (same line)

middleware_test.go
lavoiesl: Good catch

lavoiesl:
Yes, your understanding is correct. Same goes for the `NotFoundHandler`.

elithrar:
OK. Can you also update https://github.com/gorilla/mux#middleware in the docs to describe when Middleware will run / not run? I think this will clear up expectations.

A small table might work to explain it - route match vs. method match vs. no match with a "Y" or "N" for the state (or tick / cross emoji if you prefer, not fussed!).

I would argue that middleware running on *every* request is a sensible use-case though, and I think the adoption of the middleware logic without explicitly agreeing on when it runs has led us to ambiguous "fixes" like this.

lavoiesl:
The table is pretty simple, so I went for a long form explanation with an example.

README.md
```diff
@@ -491,6 +491,24 @@ r.Use(amw.Middleware)
 
 Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to. Middlewares _should_ write to `ResponseWriter` if they _are_ going to terminate the request, and they _should not_ write to `ResponseWriter` if they _are not_ going to terminate it.
 
+#### Middlewares on route-not-found
+
+Middlwares only run when a route is found. If the request can't be matched, including if the method is not allowed, the middlewares will be skipped.
```
elithrar: Middleware

README.md
```diff
@@ -491,6 +491,24 @@ r.Use(amw.Middleware)
 
 Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to. Middlewares _should_ write to `ResponseWriter` if they _are_ going to terminate the request, and they _should not_ write to `ResponseWriter` if they _are not_ going to terminate it.
 
+#### Middlewares on route-not-found
```
elithrar: Middleware and Routing

README.md
```diff
@@ -491,6 +491,24 @@ r.Use(amw.Middleware)
...
+
+Middlwares only run when a route is found. If the request can't be matched, including if the method is not allowed, the middlewares will be skipped.
+
+This is also true if you specify a custom `MethodNotAllowedHandler` or `NotFoundHandler`.
```
elithrar: : middleware will not run before/after these handlers.

kisielk:
I'm ok to merge this after the doc changes.

lavoiesl:
All good

lavoiesl:
Anything else required here?

elithrar:
OK, so I really wanted to dwell on this for a while.

- I think only running middleware, by default, on a route match is a mistake: it's non-obvious, and breaks use-cases/expectations around logging middleware, custom NotFound handlers, etc
- The path forward should be to run middleware on every request through the `Router`, or if only applied to `Subrouter`, that matches any route in the Subrouter (since applying it at the sub-router level implies a match, since a Subrouter requires a match-constraint)
- Optionally, an option on `Router` that allows you to `RunMiddlewareOnMatchOnly` for users that want to opt-in.

lavoiesl:
I tried several alternative ways, but the way the code is structured doesn't lend well to what @elithrar is asking, I couldn't find a way without very significantly rearchitecting the matching mechanisms. I already spent close to 8 hours on this and cannot afford more.

Therefore, I’m abandoning this PR for the time being.

I still very much want it and will continue to use my patch for now on, but I fear I will have to switch router at some point.

PR #390, opened 2018-07-06
Add test for .Matches(...).Subrouter()

fstanis:
Relevant issue: #389

Issue #389, opened 2018-07-05
Using Methods(...) with Subrouter() doesn't work


fstanis:
**What version of Go are you running?** (Paste the output of `go version`)
go version go1.9.4 linux/amd64

**What version of gorilla/mux are you at?** (Paste the output of `git rev-parse HEAD` inside `$GOPATH/src/github.com/gorilla/mux`)
cb4698366aa625048f3b815af6a0dea8aef9280a

**Describe your problem** (and what you have tried so far)
It seems that creating a Subrouter to filter out methods doesn't work. #300 is possibly related.

**Paste a minimal, runnable, reproduction of your issue below** (use backticks to format it)
```go
func main() {
	r := mux.NewRouter().Methods("POST").Subrouter()
	r.Path("/test").Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
	http.Handle("/", r)
	http.ListenAndServe(":8080", nil)
}
```
Navigating to http://localhost:8080/test returns an empty page.

Compare with:

```go
func main() {
	r := mux.NewRouter()
	r.Methods("POST").Path("/test").Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))
	http.Handle("/", r)
	http.ListenAndServe(":8080", nil)
}
```

...where it returns 405.

elithrar:
Thanks for the report.

Are you interested in PR’ing a failing test (first), and then we can look
to resolve?

fstanis:
Sure, created a PR for a quick minimal test, seems to failing (as expected).

PR #422, opened 2018-11-25
Subroute configuration fixes

jwilner:
## problem
In #389, @fstanis highlighted an issue with routes branched off a subrouter not filtering by `Methods()` as intended; they wrote demonstrative tests in #390. 

I identified several more matchers which suffered the same problem. Specifically, if you invoke `Schemes()`, `Headers()`, ~or `Query()`~ **(edit: this is incorrect)** before invoking `Subrouter()`, your config is discarded for child routes. I have incorporated @fstanis' tests for `Methods()` and added a few more for the other three methods; all of the negative cases (`shouldMatch: false`) would currently fail on master. 

As far as I can tell, passing on these matchers isn't addressed in the code because I don't think this is necessarily the most likely use case for `Subrouter` (I imagine it was written to effectively avoid typing out certain hierarchical path prefices over and over, and it does so very expressively) but I do think it seems worth fixing. Consider that programmers might be using `Schemes("https")` to make a best effort to ensure that all requests over a given subrouter are via https, or `Headers("User-Agent", "FooBar")` to ensure that all endpoints under the subrouter fire only for certain user agents -- neither of those use cases would currently be met.

## proposal
The fix is to make the config actually be communicated via the subrouter. The current pattern for communicating config between parent and children routes has been the `parentRoute` interface. It has its advantages, but it is verbose and decentralizes the "inheritance" logic; to adopt it here would've required adding four new methods to the interface, satisfying them on both `Router` and `Route`, and largely duplicating the logic in each of those methods.

Instead, I suggest defining a common data structure (here named `routeConf`) for those properties which need to be passed on. I used an embedded struct here to keep the relative diff as small as possible. When calling `Route` or `Subrouter`, the struct is copied and passed on to the new child object. This centralizes the inheritance logic (effectively putting it in the `copyRouteConf` func) and guarantees that the values are carried over as the API suggests.

## notes
- To clearly highlight the behavioral change in this diff, I left intact some `parentRoute` usage that is now effectively obsolete. If this PR gets a tentative 👍from the powers that be, I'd like to go ahead and strip out some of that dead code.
- The `TestWalkSubrouters` test was modified to pass with the new logic; I think that the new behavior is actually the correct behavior. The old behavior appears inconsistent to me (the second `pathTemplate` includes its parent route, but the third and fourth don't include theirs), and `GetPathTemplate()` is most likely to be useful if it provides the whole path, not a partial. If I'm misunderstanding or we want to prioritize the backwards compatibility for this scenario, I can preserve the behavior.

Thanks!

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -50,10 +50,9 @@ type Router struct {
 	// Configurable Handler to be used when the request method does not match the route.
 	MethodNotAllowedHandler http.Handler
 
-	// Parent route, if this is a subrouter.
-	parent parentRoute
 	// Routes to be matched, in order.
 	routes []*Route
+
 	// Routes by name for URL building.
 	namedRoutes map[string]*Route
 
@@ -65,6 +64,7 @@ type Router struct {
 	// Slice of middlewares to be called after a match is found
 	middlewares []middleware
 
+	// configuration shared with `Route`
 	routeConf
 }
 
@@ -82,38 +82,39 @@ type routeConf struct {
 	skipClean bool
 
 	// Manager for the variables from host and path.
-	regexp *routeRegexpGroup
+	regexp routeRegexpGroup
 
 	// List of matchers.
 	matchers []matcher
+
+	// The scheme used when building URLs.
+	buildScheme string
+
+	buildVarsFunc BuildVarsFunc
 }
 
 // returns an effective deep copy of `routeConf`
 func copyRouteConf(r routeConf) routeConf {
-	c := routeConf{
-		useEncodedPath: r.useEncodedPath,
-		strictSlash:    r.strictSlash,
-		skipClean:      r.skipClean,
+	c := r
+
+	if r.regexp.path != nil {
+		c.regexp.path = copyRouteRegexp(r.regexp.path)
 	}
 
-	if r.regexp != nil {
-		c.regexp = &routeRegexpGroup{}
-		if r.regexp.path != nil {
-			c.regexp.path = copyRouteRegexp(r.regexp.path)
-		}
-		if r.regexp.host != nil {
-			c.regexp.host = copyRouteRegexp(r.regexp.host)
-		}
-		c.regexp.queries = make([]*routeRegexp, 0, len(r.regexp.queries))
-		for _, q := range r.regexp.queries {
-			c.regexp.queries = append(c.regexp.queries, copyRouteRegexp(q))
-		}
-		c.regexp.queries = r.regexp.queries
+	if r.regexp.host != nil {
+		c.regexp.host = copyRouteRegexp(r.regexp.host)
 	}
 
+	c.regexp.queries = make([]*routeRegexp, 0, len(r.regexp.queries))
+	for _, q := range r.regexp.queries {
+		c.regexp.queries = append(c.regexp.queries, copyRouteRegexp(q))
+	}
+
+	c.matchers = make([]matcher, 0, len(r.matchers))
 	for _, m := range r.matchers {
 		c.matchers = append(c.matchers, m)
 	}
+
 	return c
 }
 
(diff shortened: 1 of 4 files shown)


elithrar:
This looks great (at first pass).

What would be useful w.r.t reviewing this is a short list of bullet points
that communicate behavior changes—such that as can use them in the release
notes verbatim as well.

jwilner:
Sure, @elithrar. I've pushed some further clean up of obsolete `parentRoute` stuff. My notes would be:

- Route matching configured via `Methods`, `Schemes`, or `Headers` now carries through `Subrouter`s to "subroutes," just as path and host matching does.
- `GetPathTemplate` always returns the path template through to the root router, whereas previously it sometimes returned an empty string.

Two edge-casey things that would only affect users chaining successive calls of methods:
- Successive calls to `BuildVarsFunc` are now always composed, whereas previously newer ones might have overridden older ones (after a subrouter). In pseudocode, `BuildVarsFunc(f).BuildVarsFunc(g) == m => g(f(m))` whereas previously `BuildVarsFunc(f).BuildVarsFunc(g) == g`.
- Built URLs now always use the first scheme argument from the latest invocation of `Schemes`,  rather than preferring the first argument of the _first_ invocation on a subroute. In other words, now `Schemes("https").Schemes("http")` will use "http" for building URLs, while previously it would've used "https" (note matching is broken in this scenario anyway since you can never satisfy both predicates).

kisielk:
@jwilner This looks great to me. I think it's a great simplification of the implementation and hopefully will help squash a lot of the bugs people have been seeing with regards to middleware and subrouters.

Thanks for taking the time to dive into the code at this level, I've been way too busy with other things and not programming with this package at the moment so the internals have been kind of unloaded from my brain :)

jwilner:
@kisielk glad it seems helpful. I did a little cursory review of the middleware issue (#378) and I don't think this solves it unfortunately. I intend to look at it more deeply though and have some ideas.

mux.go
```diff
@@ -50,24 +50,77 @@ type Router struct {
...
+	matchers []matcher
+
+	// The scheme used when building URLs.
+	buildScheme string
+
+	buildVarsFunc BuildVarsFunc
+}
+
+// returns an effective deep copy of `routeConf`
+func copyRouteConf(r routeConf) routeConf {
```
elithrar: Can you add a test for this that ensures it copies all fields? This way, any new fields added to `routeConf` will cause tests to fail if `copyRouteConf` is not updated in kind.

jwilner: Definitely, good thinking

elithrar:
Huge thanks for this @jwilner, and apologies for the delay in reviewing (been travelling for work!).

jwilner:
No worries, glad it was helpful!

tsuna:
There is no mention of this bugfix in the 1.7.0 changelog. I think it's worth mentioning because it fixes a bug that will make some user code more restrictive.  For example we have some code that does:
```go
  mux.NewRouter().Host("{customerId:[0-9]*}{dot:[.]?}example.com{hostPort:[:0-9]*}").Subrouter().StrictSlash(true).UseEncodedPath()
```

Before this change, the host matcher wasn't copied into the subrouter.  This meant that a request to `wrong.com` would pass the matchers but would fail to extract the var `customerId`, but would still be passed on to the user handler code anyway.  With this change the request no longer passes the host matcher and gets a 404 instead of going to the user handler.  To be clear, I think the new behavior is the correct one, but I'm just saying this deserves a mention in the release note as it was quite a head scratcher for me to figure out why after upgrading to 1.7.0 some tests weren't matching any route anymore.

tsuna:
In my case I had requests coming from an internal service that was using a different hostname, so I needed to change the `.Host(..)` call to `.Host({customerId:[0-9]*}{dot:[.]?}{domain:internal-name|example.com}{hostPort:[:0-9]*}")` because before v1.7.0 requests coming in with `Host: internal-name` were incorrectly passed through, while with v1.7.0 this bug has been fixed and the request will 404.

PR #423, opened 2018-12-04
transitioning to modules by first fixing this path, then removing forked

trtg:
mux altogether, then pointing back at mainline gorilla, then pinning
versions


elithrar:
I assume this was a mistake (the import path is not changing to point to another repo).

trtg:
Yup, sorry about that

Issue #429, opened 2018-12-19
router.Use middleware not being hit


ORESoftware:
**What version of Go are you running?** 
go version go1.10.4 linux/amd64

**What version of gorilla/mux are you at?**
`3d80bc801bb034e17cae38591335b3b1110f1c47`

**Describe your problem** (and what you have tried so far)
I am registering middleware, but they don't seem to be hit.

**Paste a minimal, runnable, reproduction of your issue below** (use backticks to format it)

```golang
package main

import (
	"encoding/json"
	"fmt"
	"github.com/dgrijalva/jwt-go"
	"github.com/gorilla/context"
	"github.com/mitchellh/mapstructure"
	"huru/migrations"
	"huru/models"
	"huru/models/person"
	"huru/routes"
	"net/http"
	"os"

	"github.com/gorilla/mux"
	_ "github.com/lib/pq"
	log "github.com/sirupsen/logrus"
)

func loggingMiddleware(next http.Handler) http.Handler {

	log.Println("logging middleware registered");

	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		// Do stuff here
		log.Println("Here is the request URI:",r.RequestURI)
		// Call the next handler, which can be another middleware in the chain, or the final handler.
		next.ServeHTTP(w, r)
	})
}

type Exception struct {
	Message string `json:"message"`
}

func authMiddleware(next http.Handler) http.Handler {

	log.Println("auth middleware registered");

	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {

		params := r.URL.Query()
		fmt.Println("the params are:", params);

		token, _ := jwt.Parse(params["token"][0], func(token *jwt.Token) (interface{}, error) {
			if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
				return nil, fmt.Errorf("there was an error")
			}
			return []byte("secret"), nil
		})
		claims, ok := token.Claims.(jwt.MapClaims)

		if ! (ok && token.Valid) {
			json.NewEncoder(w).Encode(Exception{Message: "Invalid authorization token"})
			return;
		}

		var user person.Model
		mapstructure.Decode(claims, &user)
		context.Set(r, "logged_in_user", user)
		next.ServeHTTP(w, r)
	})
}

func errorMiddleware(next http.Handler) http.Handler {

	log.Println("error handling middleware registered");

	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		defer func() {
[...]

PR #430, opened 2018-12-20
fix for issue #429

maplebed:
Details are in Issue #429 - @tomare suggested a patch and it seems to work for me. 

The test I added verifies that when multiple submuxes hang off a parent mux, each gets the appropriate middleware called according to the routes assigned.

### DIFF
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -173,7 +173,7 @@ func TestMultipleSubrouterMiddleware(t *testing.T) {
 	req = newRequest("GET", "/subr1")
 	router.ServeHTTP(rw, req)
 	if rmw.timesCalled != 2 {
-		t.Fatalf("Expected root middleware %d calls, but got %d", 1, rmw.timesCalled)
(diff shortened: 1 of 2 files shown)


maplebed:
hrmph. other tests fail with this change.

maplebed:
I clearly don't understand how errors are supposed to move... One of the failing tests contains this:
```go
	// Test the subrouter first
	match = new(RouteMatch)
	matched = s.Match(req, match)
	// Now we should get a match
	if !matched {
		t.Errorf("Subrouter should have matched %s", req.RequestURI)
	}
	// But MatchErr should be set to ErrNotFound anyway
	if match.MatchErr != ErrNotFound {
		t.Errorf("Expected ErrNotFound MatchErr, but was %v", match.MatchErr)
	}
```
It is asserting there _should_ be a match and that `match.MatchErr` should be `ErrNotFound` even when there was a match. grumps.

maplebed:
closing in favor of PR #431

PR #434, opened 2018-12-25
Replace domain.com with example.com

raeesbhatti:
Because domain.com is an actual business, example.com should be used for example purposes.


kisielk:
Thanks!

Issue #437, opened 2019-01-07
Status is 404 Not Found instead of 405 Method Not Allowed


g-w:
**What version of Go are you running?**

```
go version go1.11.4 linux/amd64
```

**What version of gorilla/mux are you at?**

f3ff42f93a451d7ffb2ff11cb9485f3f88089c83

**Describe your problem**

Currently the router returns `404` instead of `405` if the path matches but not the method. This is only the case if another not matching route was registered after the matching one.

This issues seems to be introduced through the changes in ef912dd76ebe9d9848c6e0fd80eaebccc9a11631 .

**Paste a minimal, runnable, reproduction of your issue below**

```
func TestMethodNotAllowed(t *testing.T) {
	router := NewRouter()
	router.HandleFunc("/thing", func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }).Methods(http.MethodGet)
	router.HandleFunc("/something", func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }).Methods(http.MethodPost)

	w := NewRecorder()
	req := newRequest(http.MethodPut, "/thing")

	router.ServeHTTP(w, req)

	if w.Code != 405 {
		t.Fatalf("Expected status code 405 (got %d)", w.Code)
	}
}
```

elithrar:
Thanks. Are you interested in submitting a PR that tackles this?

The challenge is that ef912dd and this are conflicting behaviour - you need
to track two "kinds" of matchErr (matchRouteErr, matchMethodErr?) in order
to handle this. That also won't solve host matches, etc.

g-w:
I will have a look into this and try to provide a PR.

g-w:
@elithrar I've created a PR #438.

PR #438, opened 2019-01-07
Ignore ErrNotFound while matching Subrouters

g-w:
MatchErr is set by the router to ErrNotFound if no route matches. If
no route of a Subrouter matches the error can by safely ignored. This
implementation only ignores these errors and does not ignore other
errors like ErrMethodMismatch.


kisielk:
What if you have a router whose only routes are subrouters, and all of them return ErrNotFound?

g-w:
In this case the no route should match and https://github.com/gorilla/mux/blob/master/mux.go#L167 should be reached. This sets `MatchErr` to `ErrNotFound`.

I've checked this through this test:

```
func TestMethodSubrouterNotFound(t *testing.T) {
	handler := func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusOK) }
	router := NewRouter()
	router.Path("/a").Subrouter().HandleFunc("/thing", handler).Methods(http.MethodGet)
	router.Path("/b").Subrouter().HandleFunc("/something", handler).Methods(http.MethodGet)

	w := NewRecorder()
	req := newRequest(http.MethodPut, "/not-present")

	router.ServeHTTP(w, req)

	if w.Code != 404 {
		t.Fatalf("Expected status code 404 (got %d)", w.Code)
	}
}
```

If you think it useful, I will add it to my PR.

BTW: This fixes #437

kisielk:
Yes I think adding this test is useful just to verify that this and any future changes won't break this.

g-w:
I've added the test.

PR #442, opened 2019-01-25
Call WriteHeader after setting other header(s) in the example

timucingelici:
From the docs: Changing the header map after a call to WriteHeader (or
Write) has no effect unless the modified headers are
trailers.


elithrar:
Good catch, thanks!

PR #447, opened 2019-02-13
Route.Host -matching will ignore any provided port from getHost(), if…

cognusion:
… a port isn't specified in the template

In lieu of checking the template pattern on every Match request, a bool is added to the routeRegexp, and set if the routeRegexp is a host AND there is no ":" in the template. I dislike extending the type, but I'd dislike doing a string match on every single Match, even more.

All existing mux tests pass. Additionally my application-level (Convey) positive and negative tests pass. Feel free to comment: You know your code better than I do.

```
=== RUN   TestPathHostPortUnset

  When a request is made, with a host and port specified, the correct match is found (no port specified) ✔✔✔✔✔


16 total assertions

--- PASS: TestPathHostPortUnset (0.00s)
=== RUN   TestPathHostPortSet

  When a request is made, with a host and port specified, and the port does match, a match is found ✔✔✔✔✔


21 total assertions

--- PASS: TestPathHostPortSet (0.00s)
=== RUN   TestPathHostPortSetNope

  When a request is made, with a host and port specified, and the port does not match, a match is NOT found ✔✔✔


24 total assertions

--- PASS: TestPathHostPortSetNope (0.00s)

```

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -113,6 +113,13 @@ func newRouteRegexp(tpl string, typ regexpType, options routeRegexpOptions) (*ro
 	if typ != regexpTypePrefix {
 		pattern.WriteByte('$')
 	}
+
+	var wildcardHostPort bool
+	if typ == regexpTypeHost {
+		if !strings.Contains(pattern.String(), ":") {
+			wildcardHostPort = true
+		}
+	}
 	reverse.WriteString(raw)
 	if endSlash {
 		reverse.WriteByte('/')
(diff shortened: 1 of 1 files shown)


yogo1212:
works for me

elithrar:
Sorry, I missed this: can you show application code before/after that explains this better?

I understand _what_ the change does, but not why.

yogo1212:
@elithrar 
Without this change, `r.Host("test.host.com")` won't match requests with the `Host: test.host.com:1234`.
In order to make that case work, this PR assumes that, unless a port specified (e.g. `r.Host("test.host.com:1234")`), any port is allowed to match.

cognusion:
@elithrar ,

As @yogo1212 properly described, a previous PR #383 horrifically broke a good swath of the Gorilla-using Internet by deciding that if you didn't specify a port in the route, you can't specify a port on the Request and have it match. This PR allows that PR to exist in-spirit, without the awful **requirement** of specifying ports for **every** route. Without either reverted #383 or implementing this patch if your route is r.Host("a.b.c") then http://a.b.c:80/ won't work but http://a.b.c/ will.  With this patch, unless you specify the port on the route, it assumes you don't give a flying leap what port the request tacks on.

elithrar:
Understood - #383 broke more than we anticipated (users should pin + test deps, but we also need to do a better job of checking impact).

Reviewing the PR today and can (hopefully) merge + tag a new version tonight.

elithrar:
@cognusion See https://github.com/gorilla/mux/releases/tag/v1.7.2

yogo1212:
@elithrar thank you for the notice <3

PR #459, opened 2019-03-04
CORSMethodMiddleware removing unexpected behaviour

Sapphire64:
I got a feeling that CORSMethodMiddleware got stuck between several iterations, and current condition does not make much sense to me.

See for yourself:
- If you don't specify `.Methods()` it does not work at all. That was not documented in the actual body of the middleware until this PR and it's actually needs effort to discover.
- If you do specify `.Methods("GET, "POST")` it starts to work and return the headers, but for some reason returns that you also support `OPTIONS`, which you don't, because all OPTIONS calls end up with 405 response and no headers are being set.
- If you do specify that your endpoint support OPTIONS with `.Methods("GET", "POST", "OPTIONS")` the middleware outputs something weird back to the users: `"Access-Control-Allow-Methods": "GET,POST,OPTIONS,OPTIONS"`.

I've seen that @pierreprinetti was bringing this problem before (https://github.com/gorilla/mux/issues/381, https://github.com/gorilla/mux/pull/382), and the fix was not merged for different reasons and different ideas on how things could work in Mux in the future, but since that time things have changed, and only the Middleware still keeps behaving unexpectedly and weirdly both for end users and for developers.

So without the second-commit fix the tests would fail like this:
```
Expected Access-Control-Allow-Methods 'POST,PUT,GET,OPTIONS', 
found 'POST,PUT,GET,OPTIONS,OPTIONS'
FAIL
```

I've also improved the test to remove unnecessary complexity of different similar conditions testing and actually testing out more things.

Looking forward to hear what do you think about this change!

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -33,9 +33,11 @@ func (r *Router) useInterface(mw middleware) {
 }
 
 // CORSMethodMiddleware sets the Access-Control-Allow-Methods response header
-// on a request, by matching routes based only on paths. It also handles
-// OPTIONS requests, by settings Access-Control-Allow-Methods, and then
-// returning without calling the next http handler.
+// on a request, by matching routes based only on paths. The headers will NOT be
+// returned if the handlers do not explicitly specify the list of supported methods.
+// This middleware also handles OPTIONS requests if allowed by the handler methods,
+// by settings Access-Control-Allow-Methods, and then returning without calling
+// the next http handler.
 func CORSMethodMiddleware(r *Router) MiddlewareFunc {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
@@ -59,7 +61,7 @@ func CORSMethodMiddleware(r *Router) MiddlewareFunc {
 			})
 
 			if err == nil {
-				w.Header().Set("Access-Control-Allow-Methods", strings.Join(append(allMethods, "OPTIONS"), ","))
+				w.Header().Set("Access-Control-Allow-Methods", strings.Join(allMethods, ","))
 
 				if req.Method == "OPTIONS" {
 					return
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -343,10 +343,13 @@ func TestCORSMethodMiddleware(t *testing.T) {
 	cases := []struct {
 		requestMethod          string
 		methods                []string
+		expectedResponseCode   int
 		expectedResponse       string
 		expectedAllowedMethods string
 	}{
-		{"POST", []string{"POST", "PUT", "GET"}, responseBody, "POST,PUT,GET,OPTIONS"},
+		{"POST", []string{"POST", "PUT", "GET"}, 200, responseBody, "POST,PUT,GET"},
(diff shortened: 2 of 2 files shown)


elithrar:
Thanks Anton - will take a look.

(I think adding the CORSMethodMiddleware was originally a mistake, but here
we are, so need to fix!)

Sapphire64:
@elithrar yeah, especially I understand you now even more than when I made this PR, this particular code is not compatible much with https://github.com/gorilla/handlers/blob/master/cors.go and is not solving any CORS challenges. It probably needs to be integrated into gorilla/handlers code or just removed completely.

elithrar:
The other part: having to specify the methods via .Methods was an
intentional decision as otherwise you have to assume _all_ methods, which
may not be the case.

I’m not sure implicitly dumping all HTTP methods into the ACAM response
header makes sense otherwise.

One thing we can do: clearer, more obvious docs. I’m going to review your
PR this weekend (been traveling all week), but README and godoc additions
to clarify usage are always my first places to look when reviewing -
especially so as mux has grown 😉

Sapphire64:
@elithrar hi Matt, didn't have much time to come back to this PR yet, I think for major change it could be combined with/replaced by handlers CORS logic, but I need to find time to look more into this. Seems that you've been busy too :)

fharding1:
Hey @Sapphire64. I agree that the current behavior is pretty terrible. I wrote the first iteration of `CORSMethodMiddleware` when I was just getting into Go and web development, and looking back I have no clue what I was thinking. While I think the _idea_ is nice: automatically set the `Access-Control-Allow-Methods` headers based on `.Methods()` so you end up with accurate CORS headers (unlike some APIs that just set that header to pretty much all methods). The implementation is _abysmal_ and practically unusable in it's current state.

@elithrar said adding it in the first place was somewhat of a mistake, and I have to say that I agree. I wouldn't be against completely removing it.

That being said, I opened a PR to fix the issues you outlined here (I actually didn't see this PR before opening here) as well as some other ones. Would appreciate if you could take a look if you're not too busy: https://github.com/gorilla/mux/pull/477

Sapphire64:
@fharding1 thanks, I'll take a look this week! Have you counted in the fact of https://github.com/gorilla/handlers/blob/master/cors.go existence? I think we ended up using it for most of the things.

fharding1:
@Sapphire64 Sweet, thanks! I have seen `handlers`.I guess you can set `Access-Control-Allow-Headers` per route with `handlers` currently, it will just be tedious. I previously thought it wasn't possible to implement this "automatic" functionality with the public `mux` API, [however I think it actually is possible](https://play.golang.org/p/ZTmIS7wZmBR). In that case, we could implement this functionality there and remove it from `mux` completely.

@elithrar "(I think adding the CORSMethodMiddleware was originally a mistake, but here
we are, so need to fix!)" do you think we need to support `CORSMethodMiddleware`, or could it be completely removed? It's fairly safe to assume nobody uses it now, considering it's almost totally broken.

elithrar:
See #477 - and thanks for motivating us to fix this, @Sapphire64.

Issue #460, opened 2019-03-13
SkipClean does not work on Subrouters or any subroutes


gavbaa:
**What version of Go are you running?** (Paste the output of `go version`)
go version go1.11.5 darwin/amd64

**What version of gorilla/mux are you at?** (Paste the output of `git rev-parse HEAD` inside `$GOPATH/src/github.com/gorilla/mux`)
15a353a636720571d19e37b34a14499c3afa9991

**Describe your problem** (and what you have tried so far)
SkipClean when applied only on subrouters does not check for cleaned URLs on subroutes.  It only applies to the entire routing event, on or off.

What we expect to happen: On routes that are declared with SkipClean(false), those routes are checked against the cleaned URL.  On routes that are declared with SkipClean(true), those routes are checked against the original URL.

Or else, SkipClean should only be allowed on the parent Router, and not on any subrouters, since it's currently ineffective.

(This probably opens the door for also processing those routes as an internal redirect instead of a public 302 as an option, but I think that should be a separate ticket and isn't required for this behavior).

**Paste a minimal, runnable, reproduction of your issue below** (use backticks to format it)
```
func TestSkipCleanSubrouter(t *testing.T) {
	func1 := func(w http.ResponseWriter, r *http.Request) {}
	func2 := func(w http.ResponseWriter, r *http.Request) {}

	r := NewRouter()
	r.HandleFunc("/api/", func1).Name("func2")
	r.SkipClean(true)
	subRouter := r.PathPrefix("/v0").Subrouter().SkipClean(false)
	subRouter.HandleFunc("/action/do/", func2).Name("func2")

	req, _ := http.NewRequest("GET", "http://localhost/api/?abc=def", nil)
	res := NewRecorder()
	r.ServeHTTP(res, req)

	if len(res.HeaderMap["Location"]) != 0 {
		t.Errorf("Req 1: Shouldn't redirect since route is already clean")
	}

	req, _ = http.NewRequest("GET", "http://localhost//api/?abc=def", nil)
	res = NewRecorder()
	r.ServeHTTP(res, req)

	if len(res.HeaderMap["Location"]) != 0 {
		t.Errorf("Req 2: Shouldn't redirect since skip clean is disabled")
	}

	req, _ = http.NewRequest("GET", "http://localhost/v0/action//do/?ghi=jkl", nil)
	res = NewRecorder()
	r.ServeHTTP(res, req)

	if len(res.HeaderMap["Location"]) == 0 {
[...]

PR #463, opened 2019-03-21
Allow SkipClean to be evaluated on each route independently.

gavbaa:
This provides a solution to #460 .  Cleaning is now moved into the route matcher, and is only activated if SkipClean has been set for that specific router (whether explicitly, or by inheritance from a parent route upon creation).

Note that due to the lack of parent knowledge in a Route/Router (which was removed a few months ago), SkipClean() changes only apply to routes that were created after the SkipClean change.  Since we have no handle or reference to a Subrouter() once it's been created from the parent router (it's only stored as a `matcher` interface), we can't make SkipClean() trickle down to the already existing children.  New ones will pick it up though.  Since this is how other current properties behave, this seemed to be expected behavior.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -116,6 +116,8 @@ func copyRouteConf(r routeConf) routeConf {
 		c.matchers = append(c.matchers, m)
 	}
 
+	c.skipClean = r.skipClean
+
 	return c
 }
 
@@ -173,26 +175,6 @@ func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 // When there is a match, the route variables can be retrieved calling
 // mux.Vars(request).
 func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
-	if !r.skipClean {
-		path := req.URL.Path
-		if r.useEncodedPath {
-			path = req.URL.EscapedPath()
-		}
-		// Clean path to canonical form and redirect.
-		if p := cleanPath(path); p != path {
-
-			// Added 3 lines (Philip Schlump) - It was dropping the query string and #whatever from query.
-			// This matches with fix in go 1.2 r.c. 4 for same problem.  Go Issue:
-			// http://code.google.com/p/go/issues/detail?id=5252
-			url := *req.URL
-			url.Path = p
-			p = url.String()
-
-			w.Header().Set("Location", p)
-			w.WriteHeader(http.StatusMovedPermanently)
-			return
-		}
-	}
 	var match RouteMatch
 	var handler http.Handler
 	if r.Match(req, &match) {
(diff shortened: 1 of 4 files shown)


elithrar:
Thank you. I’m really behind in review backlog, but a quick ask: also
document the “existing children” behavior in the godoc for SkipClean?

Bonus points for adding an example to the README too.

mux.go
```diff
@@ -413,6 +406,13 @@ type RouteMatch struct {
 	Handler http.Handler
 	Vars    map[string]string
 
+	// Cleaned version of the path being matched.
```
elithrar: Since these are exported, can you provide clearer guidance for a user trying to understand either:

* What their value is?
* Should I be setting this?
* What the default is?

regexp.go
```diff
@@ -170,7 +171,23 @@ func (r *routeRegexp) Match(req *http.Request, match *RouteMatch) bool {
 		if r.options.useEncodedPath {
 			path = req.URL.EscapedPath()
 		}
-		return r.regexp.MatchString(path)
+		if !r.options.skipClean {
```
elithrar: Please document this code-path more clearly - e.g. "why".

elithrar:
Hey @gavbaa - are you still interested in finishing this one off?

gavbaa:
Yes, just have to set some time aside to go through and address the comments.

theverything:
Is there any way that I can help to get this merged? I need this!!

fharding1:
@theverything you could create your own branch and complete the remainder of the work if you'd like.

mux.go
```diff
@@ -209,6 +191,17 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 	}
 
+	if match.OnlyMatchedCleanPath {
+		// Added 3 lines (Philip Schlump) - It was dropping the query string and #whatever from query.
+		// This matches with fix in go 1.2 r.c. 4 for same problem.  Go Issue:
+		// http://code.google.com/p/go/issues/detail?id=5252
+		url := *req.URL
+		url.Path = match.CleanPath
+		w.Header().Set("Location", url.String())
+		w.WriteHeader(http.StatusMovedPermanently)
```
JoelSpeed, line 201: Have you considered using `http.Redirect` here instead? Would be more consistent with what other projects are doing.
Basically the difference is that it would give a response body with the location in too and a moved permanently message.

JoelSpeed (commented):
Started looking into this today and realised the documentation for `SkipClean` isn't correct at the moment, but this PR seems like it should fix that. Is there anything I can do to help get this PR across the finish line?

amustaque97:
Hey @gavbaa, @theverything  - are you guys working on this PR?

mux.go
jarri-abidi, line 201:
```suggestion
		http.Redirect(w, req, url.String(), http.StatusMovedPermanently
```
based on @JoelSpeed 's suggestion

jaitaiwan:
Just noting no response as yet - will plan on coming back around to this if there's further updates.

mobyvb:
@jaitaiwan are you still working on this?

Issue #420, opened 2018-11-16
Use 308 instead of 301 for trailing slash at least for non-GET methods


or-else:
This is a new issue to bring back the discussion of #285 

It seems like it has to be done at some point or at least it should be made configurable. The broken clients should not prevent newer clients from working correctly.

Maybe make it HTTP version-dependent, i.e.HTTP/ 1.0 should get 301 while HTTP/1.1 gets 308. Certainly HTTP/2.0 should get 308.

elithrar:
Earmarking this for v2.0, but won’t change today.

You can use middleware to change 301/2’s into 307/8’s in the meantime.

or-else:
Thanks!

I can sure find a workaround, the simplest being just to make sure all URLs have a trailing slash. 

It's that I just killed half a day trying to figure out why Android clients can't upload.
And the possible fix on Gorilla's side seems to be so trivial.

drum445:
This would be a nice change, would mean I can delete my clunky middleware I am using currently

elithrar:
I'd accept a PR that makes this an option on the Router, but am unlikely to change this as a default.

PR #465, opened 2019-04-03
knob to set clean url with a 308 redirect

toravir:
fixes issue described in #420 

I am a little unsure if there should be a check to see if the method is non-GET and then only do this or is it good to do this for everything (current pr) - please comment.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -92,6 +92,11 @@ type routeConf struct {
 	buildScheme string
 
 	buildVarsFunc BuildVarsFunc
+
+	// If true, if the url is cleaned, will send
+	// HTTP Status code 308 (Permanent Redirect) instead of
+	// the default 301 (Moved Permanently)
+	cleanRedirectUse308 bool
 }
 
 // returns an effective deep copy of `routeConf`
@@ -189,7 +194,11 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			p = url.String()
 
 			w.Header().Set("Location", p)
-			w.WriteHeader(http.StatusMovedPermanently)
+			if r.cleanRedirectUse308 {
+				w.WriteHeader(http.StatusPermanentRedirect)
+			} else {
+				w.WriteHeader(http.StatusMovedPermanently)
+			}
 			return
 		}
 	}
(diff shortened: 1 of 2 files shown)


toravir:
Just noticed that https://github.com/gorilla/mux/pull/463 conflicts with this PR.. will rework this PR if the other one is merged first.

mux.go
```diff
@@ -92,6 +92,11 @@ type routeConf struct {
 	buildScheme string
 
 	buildVarsFunc BuildVarsFunc
+
+	// If true, if the url is cleaned, will send
```
elithrar: RE: this - 

* How do you know the URL is cleaned? (note: uppercase "URL")
* Why is this useful to a user to enable, if they haven't read this PR or the history?

mux.go
```diff
@@ -363,6 +372,15 @@ func (r *Router) Walk(walkFn WalkFunc) error {
 	return r.walk(walkFn, []*Route{})
 }
 
+// CleanRedirectUse308 defines the redirect code following a path cleaning. The
+// default value is false - meaning it will use 301 (Moved Permanently) code.
```
elithrar: As a user, what would a 308 get me over a 301?

(Note: *I* know, but someone trying to learn mux will not)

toravir:
> It's not clear that this is a useful addition for _most_ users. Those who really desire to use 308's can wrap the ResponseWriter with middleware.

Middleware is simple for (301->308) - but it becomes complicated if it needs to be done when URL cleaning was done.. wrote up one: https://gist.github.com/toravir/8c9bab5756f33963fa745d977d85f94e - but still it does not handle/relay other status codes - then there is the 'multiple response.WriteHeader calls' error if middleware tries to overwrite the status. Agree, most users will not need this - but few who want to do the 308, this knob provides a simple way to do. Will address comments about the docs later in the day..

elithrar:
You need to handle the 'wrote header' case - the standard library implementations of `http.ResponseWriter` often have a `wroteHeader bool` field: https://golang.org/src/net/http/server.go

Working example that rewrites 301/302 and wraps the entire router:

```go
package main

import (
	"fmt"
	"log"
	"net/http"

	"github.com/gorilla/mux"
)

type statusResponseWriter struct {
	http.ResponseWriter
	wroteHeader bool
}

func (srw *statusResponseWriter) WriteHeader(code int) {
	if !srw.wroteHeader {
		switch code {
		case 301:
			code = 308
		case 302:
			code = 307
		}

		srw.wroteHeader = true
		srw.ResponseWriter.WriteHeader(code)
	}
}

func statusRewriter(next http.Handler) http.Handler {
	fn := func(w http.ResponseWriter, r *http.Request) {
		wrappedWriter := &statusResponseWriter{ResponseWriter: w}
		next.ServeHTTP(wrappedWriter, r)
	}

	return http.HandlerFunc(fn)
}

func handler(w http.ResponseWriter, r *http.Request) {
	http.Redirect(w, r, "/target", 302)
}

func target(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintln(w, "hello")
}

func main() {
	r := mux.NewRouter()

	r.HandleFunc("/", handler)

	log.Fatal(
		http.ListenAndServe("localhost:8000",
			statusRewriter(r),
		))
}
```

toravir:
Ah.. translate the status at the first writing itself.. yes that does the work. Earlier, I was only thinking of re-writing the status after the actual handler completes.

This knob isn't useful - unless in a corner case where one wants a behavior of using 308 ONLY for the case of URL cleaning is done. In my current project, i don't have any handlers returning 301 - so i can use the blanket middleware to do the job - since 301 is generated only when url is cleaned.

Thx. Will close the PR in a day.

elithrar:
Yep! The middleware doesn’t “care”.

I’ll add the middleware example to the docs to make it easier for the next
person.

PR #474, opened 2019-05-15
Guess the scheme if r.URL.Scheme is unset

euank:
It's not expected that the request's URL is fully populated when used on
the server-side (it's more of a client-side field), so we shouldn't
expect it to be present.

In practice, it's only rarely set at all on the server, making mux's
`Schemes` matcher tricky to use as it is.

This commit adds a test which would have failed before demonstrating the
problem, as well as a fix which I think makes `.Schemes` match what
users expect.

This will technically break anyone who did the following:

```go
router.Schemes("") // match http and https
```

### DIFF
--- a/mux_httpserver_test.go
+++ b/mux_httpserver_test.go
@@ -1,0 +1,46 @@
+// +build go1.9
+
+package mux
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestSchemeMatchers(t *testing.T) {
+	router := NewRouter()
+	router.HandleFunc("/", func(rw http.ResponseWriter, r *http.Request) {
+		rw.Write([]byte("hello world"))
+	}).Schemes("http", "https")
+
+	assertHelloWorldResponse := func(t *testing.T, s *httptest.Server) {
+		resp, err := s.Client().Get(s.URL)
+		if err != nil {
+			t.Fatalf("unexpected error getting from server: %v", err)
+		}
+		if resp.StatusCode != 200 {
+			t.Fatalf("expected a status code of 200, got %v", resp.StatusCode)
+		}
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatalf("unexpected error reading body: %v", err)
+		}
+		if !bytes.Equal(body, []byte("hello world")) {
+			t.Fatalf("response should be hello world, was: %q", string(body))
+		}
+	}
+
+	t.Run("httpServer", func(t *testing.T) {
+		s := httptest.NewServer(router)
+		defer s.Close()
+		assertHelloWorldResponse(t, s)
+	})
+	t.Run("httpsServer", func(t *testing.T) {
+		s := httptest.NewTLSServer(router)
+		defer s.Close()
+		assertHelloWorldResponse(t, s)
+	})
+}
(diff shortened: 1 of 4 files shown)


route.go
```diff
@@ -412,7 +412,18 @@ func (r *Route) Queries(pairs ...string) *Route {
...
 
 func (m schemeMatcher) Match(r *http.Request, match *RouteMatch) bool {
-	return matchInArray(m, r.URL.Scheme)
+	scheme := r.URL.Scheme
+	// https://golang.org/pkg/net/http/#Request
+	// "For [most] server requests, fields other than Path and RawQuery will be
+	// empty."
+	// Since we're an http muxer, the scheme is either going to be http or https
+	// though, so we can just set it based on the tls termination state.
+	if r.URL.Scheme == "" && r.TLS == nil {
```
betawaffle: How about replacing `r.URL.Scheme` in these checks with `scheme` (the variable you define above)? Something like this:

```go
if scheme == "" {
	if r.TLS == nil {
		scheme = "http"
	} else {
		scheme = "https"
	}
}
```

fharding1:
To back up `it's only rarely set at all on the server`:

https://golang.org/pkg/net/http/#Request
```
        // URL specifies either the URI being requested (for server
        // requests) or the URL to access (for client requests).
        //
        // For server requests, the URL is parsed from the URI
        // supplied on the Request-Line as stored in RequestURI.  For
        // most requests, fields other than Path and RawQuery will be
        // empty. (See RFC 7230, Section 5.3)
        //
        // For client requests, the URL's Host specifies the server to
        // connect to, while the Request's Host field optionally
        // specifies the Host header value to send in the HTTP
        // request.
        URL *url.URL
```

euank:
I added some docs to Route.Schemes. Since `Router` just references the `Route` docs, it seems like that's the only place docs should be updated.

route.go
```diff
@@ -412,11 +412,30 @@ func (r *Route) Queries(pairs ...string) *Route {
...
+	// https://golang.org/pkg/net/http/#Request
+	// "For [most] server requests, fields other than Path and RawQuery will be
+	// empty."
+	// Since we're an http muxer, the scheme is either going to be http or https
+	// though, so we can just set it based on the tls termination state.
+	if scheme == "" {
+		if r.TLS == nil {
+			scheme = "http"
+		} else {
+			scheme = "http"
```
jameshalsall:
```suggestion
			scheme = "https"
```

euank: Thanks for catching that.

euank:
I updated the test so it would have caught that typo and rebased it on master to pickup the CI config changes.

elithrar:
@fharding1 - PTAL.

mux_httpserver_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestSchemeMatchers(t *testing.T) {
```
fharding1: Would it make sense for this test to have different responses for the different schemes so we can use that to differentiate? I believe this test would pass if the httpsServer subtest was matching on the http route.

elithrar:
Thanks for taking a look - agree that the test cases should be split out
here.

mux_httpserver_test.go
euank: I think you're right. I made it so there's one router that has both matchers, and that each matcher has a different response. I think that should better test for what we care about.
Thanks for the review!

fharding1:
@euank I'll take another look at this today

elithrar:
Thanks both!

PR #477, opened 2019-05-27
Improve CORS Method Middleware

### LINKED ISSUE #476: Make CORSMethodMiddleware actually make sense
I was the one that added `CORSMethodMiddleware`, and I feel bad about it looking back, because it's implementation is terrible. I didn't understand enough about CORS—just had some half-baked idea about something that might be cool and decided to PR it. It's understandably [caused confusion](https://github.com/gorilla/mux/issues/381). I would like to change it in the following ways so it's actually useful:

* Only set the header if an `OPTIONS` method matcher exists for the route
* Set `Access-Control-Allow-Methods` on all requests (not just preflight, see https://stackoverflow.com/questions/24264574/cors-headers-present-only-on-preflight-or-every-request)
* Do not return after setting the header. Since people need to create their own `OPTIONS` handlers for this middleware to work at all, and they might want to do other things in the `OPTIONS` handler (such as set other CORS headers), it doesn't make sense to return
* Do not append `OPTIONS` to the allowed methods. This is currently a bug. If there isn't an `OPTIONS` matcher the middleware will only set the header on non-`OPTIONS` requests (which is invalid) and it will say that there is an `OPTIONS` method when there isn't. If there is an `OPTIONS` matcher it will return an `Access-Control-Allow-Methods` with two `OPTIONS` (not really invalid, but bad)
* Vastly improve the documentation to document all this behavior
* Add more (better documented) tests
* Add an example test

I believe making these changes would both make the middleware more intuitive and practical to use. It is a fairly breaking change but I think it's worth it since the current behavior is practically useless.

fharding1:
This PR implements the behavior outlined in #476:

* Sets `Access-Control-Allow-Methods` on all requests to a route with an OPTIONS method matcher
* Does not return after setting the `Access-Control-Allow-Methods` header
* Does not append `OPTIONS` header to `Access-Control-Allow-Methods` regardless of whether there is an OPTIONS method matcher. Only returns `OPTIONS` if there is an actual matcher
* Adds tests for the above behavior

Closes #476


fharding1:
It looks like CI is failing on Go < 1.11 due to `diff -u <(echo -n) <(gofmt -d .)`. Likely related to https://github.com/golang/go/issues/26228. Not sure what I should do about this. Should CI be modified to only check the `gofmt` diff on Go >= 1.11?

middleware.go
```diff
@@ -32,41 +32,60 @@ func (r *Router) useInterface(mw middleware) {
 	r.middlewares = append(r.middlewares, mw)
 }
 
-// CORSMethodMiddleware sets the Access-Control-Allow-Methods response header
-// on a request, by matching routes based only on paths. It also handles
-// OPTIONS requests, by settings Access-Control-Allow-Methods, and then
-// returning without calling the next http handler.
+// CORSMethodMiddleware automatically sets the Access-Control-Allow-Methods response header
+// on preflight requests for routes that have an OPTIONS method matcher to all the method
```
fharding1: I just realized this behavior is wrong. CORS headers should be set on all requests, including non-preflight.

fharding1:
I am closing this until a decision is reached on what to do from the conversation in #459

fharding1:
I've modified the travis CI config to only check the gofmt diff on the latest Go version, let me know if that is undesirable for some reason.

elithrar:
Good move - no disagreement from me.

Will review in full this week.

elithrar:
Can you summarize the breaking changes / major behavioral changes (same thing) for the change-log, too?

fharding1:
Sure! I think these are all of them:

Breaking changes to `CORSMethodMiddleware`:
* Only sets the `Access-Control-Allow-Methods` header if an `OPTIONS` method matcher exists for the route
* Sets `Access-Control-Allow-Methods` on all requests to a route using  `CORSMethodMiddleware` that have an `OPTIONS` method matcher, not just `OPTIONS`/preflight requests
* Does not return after setting the header, continues to request handler
* Does not append `OPTIONS` to allowed methods. The header will be set to the list of method matchers for a route, that's all

.travis.yml
```diff
@@ -19,6 +19,9 @@ install:
...
 script:
   - go get -t -v ./...
-  - diff -u <(echo -n) <(gofmt -d .)
-  - if [[ "$LATEST" = true ]]; then go vet .; fi
+  - |
+    if [[ "$LATEST" = true ]]; then
+      diff -u <(echo -n) <(gofmt -d .)
+      go vet .
+    fi
   - go test -v -race ./...
```
elithrar: Can you rebase this against master - it will remove this

README.md
```diff
@@ -29,6 +29,7 @@ The name mux stands for "HTTP request multiplexer". Like the standard `http.Serv
 * [Walking Routes](#walking-routes)
 * [Graceful Shutdown](#graceful-shutdown)
 * [Middleware](#middleware)
+* [CORS Method Middleware](#corsmethodmiddleware)
```
elithrar: "Handling CORS Requests" might make more sense as a title/index

README.md
```diff
@@ -491,6 +492,62 @@ r.Use(amw.Middleware)
 
 Note: The handler chain will be stopped if your middleware doesn't call `next.ServeHTTP()` with the corresponding parameters. This can be used to abort a request if the middleware writer wants to. Middlewares _should_ write to `ResponseWriter` if they _are_ going to terminate the request, and they _should not_ write to `ResponseWriter` if they _are not_ going to terminate it.
 
+### CORSMethodMiddleware
```
elithrar: Same as above w.r.t title

README.md
```diff
@@ -491,6 +492,62 @@ r.Use(amw.Middleware)
...
+
+`CORSMethodMiddleware` intends to make it easier to strictly set the `Access-Control-Allow-Methods` response header. You will still need to use your own CORS handler to set the other CORS headers such as `Access-Control-Allow-Origin`.
```
elithrar: Link to the godoc for it as well - https://godoc.org/github.com/gorilla/mux#CORSMethodMiddleware

README.md
```diff
@@ -491,6 +492,62 @@ r.Use(amw.Middleware)
...
+
+`CORSMethodMiddleware` will set the `Access-Control-Allow-Methods` header to all the method matchers (e.g. `r.Methods(http.MethodGet, http.MethodPut, http.MethodOptions)` -> `Access-Control-Allow-Methods: GET,PUT,OPTIONS`) on a route. Note that there must be an `OPTIONS` method matcher for the middleware to set the headers.
```
elithrar: I would break this out into bullets to make it more readable.

* You will still need to use your own CORS handler to set the other CORS headers such as `Access-Control-Allow-Origin` - e.g. such as [rs/cors](https://github.com/rs/cors)
* The middleware will set the `Access-Control-Allow-Methods` header to all the method matchers (e.g. `r.Methods(http.MethodGet, http.MethodPut, http.MethodOptions)` -> `Access-Control-Allow-Methods: GET,PUT,OPTIONS`) on a route. 
* If you do not specify any methods, then <behaviour>

> **Important**: there must be an `OPTIONS` method matcher - `.Methods(http.MethodOptions)` - for the middleware to set the CORS response headers.

README.md
```diff
@@ -491,6 +492,62 @@ r.Use(amw.Middleware)
...
+package main
+
+import (
+	"net/http"
+	"github.com/gorilla/mux"
+)
+
+func main() {
+    r := mux.NewRouter()
+    r.HandleFunc("/foo", fooHandler).Methods(http.MethodGet, http.MethodPut, http.MethodPatch, http.MethodOptions)
```
elithrar: Add a comment above reinforcing the need to specify the `MethodOptions` method, etc.

elithrar: (it keeps getting missed, changed, etc - make it crystal clear for users!)

README.md
```diff
@@ -491,6 +492,62 @@ r.Use(amw.Middleware)
...
+    w.Header().Set("Access-Control-Allow-Origin", "*")
+    if r.Method == http.MethodOptions {
+        return
+    }
+
+    w.Write([]byte("foo"))
+}
+```
+
+And an `OPTIONS` request to `/foo` would look like:
```
elithrar: Describe how a user can test. Many keep trying with `curl` and don't set an `Origin` header or CORS request headers to trigger this.

i.e. possibly show a way for a user to reproduce.

example_cors_method_middleware_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"net/http/httptest"
+
+	"github.com/gorilla/mux"
+)
+
+func ExampleCORSMethodMiddleware() {
+	r := mux.NewRouter()
+
+	r.HandleFunc("/foo", func(w http.ResponseWriter, r *http.Request) {
+		// Do something here
```
elithrar: "Handle the request"

middleware.go
```diff
@@ -32,41 +32,47 @@ func (r *Router) useInterface(mw middleware) {
 	r.middlewares = append(r.middlewares, mw)
 }
 
-// CORSMethodMiddleware sets the Access-Control-Allow-Methods response header
-// on a request, by matching routes based only on paths. It also handles
-// OPTIONS requests, by settings Access-Control-Allow-Methods, and then
-// returning without calling the next http handler.
+// CORSMethodMiddleware automatically sets the Access-Control-Allow-Methods response header
+// on requests for routes that have an OPTIONS method matcher to all the method matchers on
+// the route. See the example for usage.
```
elithrar: "Routes that do not explicitly handle OPTIONS requests will not be processed by this middleware" - or something to that effect.

README.md
fharding1: Good idea, will do.

README.md
fharding1: Yep, it should be more obvious here. Will add a comment.

.travis.yml
fharding1: I rebased, but I'm not sure what you mean about it removing this?

README.md
fharding1: They don't need to set `Origin` for CORS Method Middleware, but agree about adding the cURL command used.

example_cors_method_middleware_test.go
fharding1: I was going off the existing example for authentication middleware, but I can update this.

.travis.yml
fharding1: Alright rebased again should be deleted now

fharding1:
Updated the circleci config to only check the gofmt diff on the latest go version like I did for travis originally. Also that github integration with circleci is clean :raised_hands:

elithrar:
Thanks! I’ll update the base CI template to do that + run lint where possible too^

^ A couple of Gorilla packages fail golint due to Http vs. HTTP casing.

.circleci/config.yml
```diff
@@ -11,8 +11,11 @@ jobs:
       - checkout
       - run: go version
       - run: go get -t -v ./...
-      - run: diff -u <(echo -n) <(gofmt -d .)
-      - run: if [[ "$LATEST" = true ]]; then go vet -v .; fi
+      - run: >
+          if [[ "$LATEST" = true ]]; then
+            diff -u <(echo -n) <(gofmt -d .)
```
elithrar: Might be preferable to have as distinct steps to make it easier to identify failure?

fharding1: Good call

elithrar:
See
https://github.com/gorilla/.github/blob/master/config.yml.tmpl as the
canonical example!

elithrar:
Thanks again @fharding1! 👍

README.md
```diff
@@ -492,6 +493,73 @@ r.Use(amw.Middleware)
...
+* TCP_NODELAY set
+* Connected to localhost (::1) port 8080 (#0)
+> GET /foo HTTP/1.1
+> Host: localhost:8080
+> User-Agent: curl/7.59.0
+> Accept: */*
+> 
+< HTTP/1.1 200 OK
+< Access-Control-Allow-Methods: GET,PUT,PATCH,OPTIONS
+< Access-Control-Allow-Origin: *
```
sam-github: The docs added above said

> You will still need to use your own CORS handler to set the other CORS headers such as `Access-Control-Allow-Origin`

But the example shows that header is added, which is it?

And what's the relationship of this middleware with github.com/gorilla/handlers/CORS?

I'm confused.

PR #478, opened 2019-05-27
Use subtests for middleware tests

fharding1:
Part of #409, awaiting feedback on whether I should update `old_test.go` before tagging that this completely closes that issue.

Uses Go subtests for all the middleware tests. I don't love the solution for `CORSMethodMiddleware`, because those tests don't really make sense, because that middleware doesn't really make sense (my bad, see #476). The new tests I wrote in #477 use subtests and completely replace the current tests for that, though.

### DIFF
--- a/middleware_test.go
+++ b/middleware_test.go
@@ -25,14 +25,24 @@ func TestMiddlewareAdd(t *testing.T) {
 	router := NewRouter()
 	router.HandleFunc("/", dummyHandler).Methods("GET")
 
-	for i := 0; i < 3; i++ {
-		t.Run(fmt.Sprintf("adds %d middlewares", i+1), func(t *testing.T) {
-			mw := &testMiddleware{}
-			router.useInterface(mw)
-			if len(router.middlewares) != i+1 || router.middlewares[i] != mw {
-				t.Fatalf("Middleware %d was not added correctly", i+1)
-			}
-		})
+	mw := &testMiddleware{}
+
+	router.useInterface(mw)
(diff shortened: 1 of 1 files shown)


middleware_test.go
```diff
@@ -24,24 +25,14 @@ func TestMiddlewareAdd(t *testing.T) {
...
-	}
-
-	banalMw := func(handler http.Handler) http.Handler {
-		return handler
-	}
-	router.Use(banalMw)
-	if len(router.middlewares) != 3 {
-		t.Fatal("MiddlewareFunc method was not added correctly")
+	for i := 0; i < 3; i++ {
+		t.Run(fmt.Sprintf("adds %d middlewares", i+1), func(t *testing.T) {
```
elithrar: Describe the test better. Not sure a sub-test makes as much sense here either?

fharding1: Yeah, this one felt weird. I'll revert changing this to subtests. :+1:

elithrar:
+ can you rebase this against master so CI will fire? 

Else I’ll push a rebased branch over the top of this later today!

fharding1:
@elithrar rebased :ok_hand:

elithrar:
One more time please @fharding1? 🥀  I think I was missing a checkbox on the @circleci side for triggering builds from forks (which wasn't on by default).

fharding1:
Rebased :heavy_check_mark:

elithrar:
Thanks again, @fharding1!

PR #483, opened 2019-06-16
GetRouteByPath utility function

bluecloudes:
GetRouteByPath returns a route registered with the given path.
An error will be returned if the route does not exists

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -223,6 +223,17 @@ func (r *Router) GetRoute(name string) *Route {
 	return r.namedRoutes[name]
 }
 
+// GetRouteByPath returns a route registered with the given path.
+// An error will be returned if the route does not exists
+func (r *Router) GetRouteByPath(path string) (*Route, error) {
+	for _, route := range r.routes {
+		if template, err := route.GetPathTemplate(); err == nil && template == path {
+			return route, nil
+		}
+	}
+	return nil, errors.New("mux: route does not exists")
+}
+
 // StrictSlash defines the trailing slash behavior for new routes. The initial
 // value is false.
 //


mux.go
```diff
@@ -223,6 +223,17 @@ func (r *Router) GetRoute(name string) *Route {
 	return r.namedRoutes[name]
 }
 
+// GetRouteByPath returns a route registered with the given path.
+// An error will be returned if the route does not exists
+func (r *Router) GetRouteByPath(path string) (*Route, error) {
+	for _, route := range r.routes {
+		if template, err := route.GetPathTemplate(); err == nil && template == path {
```
elithrar: This doesn't do what you think it will do: the path template could be a more complex match (e.g. `/{name}/`) which could be non-obvious to someone expecting a proper match.

Can you explain the use-case here? https://godoc.org/github.com/gorilla/mux#Router.Get + naming your routes will be more precise here.

bluecloudes: My use-case is as below
1. A Router object already created, with HandlerFunc, without any name attachment
```
//like below
r := mux.NewRouter()
r.HandleFunc("/{name1}/", handlerFunc1)
r.HandleFunc( "/{name2}/", handlerFunc2)
```
2. Given router, I want to get a route for "/{name1}/" (exact match)
3. After getting the route, I want to update Handler
Any suggestions?

fharding1: Is this a common enough use-case that this PR should be integrated into the library?

bluecloudes: Currently for any route, if name is not attached, it's not possible to access that route back. Route doesn't have a default name as-well. So, I thought a GetRouteByPath would be a useful utility

fharding1: Sure, but how often do people actually need to do that? It seems like a really uncommon use-case to me, but maybe I'm missing something.

PR #484, opened 2019-06-16
[WIP] Create CircleCI config

elithrar:
Validate CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,63 @@
+version: 2.0
+
+jobs:
+  # Base test configuration for Go library tests Each distinct version should
+  # inherit this base, and override (at least) the container image used.
+  "test": &test
+    docker:
+      - image: circleci/golang:latest
+    working_directory: /go/src/github.com/gorilla/mux
+    steps: &steps
+      - checkout
+      - run: go version
+      - run: go get -t -v ./...
+      - run: diff -u <(echo -n) <(gofmt -d .)
+      - run: if [[ "$LATEST" = true ]]; then go vet -v .; fi
+      - run: go test -v -race ./...
+
+  "latest":
+    <<: *test
+    environment:
+      LATEST: true
+
+  "1.12":
+    <<: *test
+    docker:
+      - image: circleci/golang:1.12
+
+  "1.11":
+    <<: *test
+    docker:
+      - image: circleci/golang:1.11
+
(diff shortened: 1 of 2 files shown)


elithrar:
PS: Don't merge, this is a WIP / demo / validating whether it's worth migrating to.

fharding1:
My $0.02: I've used CircleCI for a few projects and have had nothing but good experiences. Simple configuration, decent build time, and used widely enough for most people to be somewhat familiar with it. Then again, I could most of those things for Travis :)

:+1: moving to CircleCI

elithrar:
Istio, et. al using it also inspire a bit of confidence. I'm going to leave this open for the time being - I also need to think about moving all Gorilla projects over (not just one) as part of this.

elithrar:
Some initial explorations around DRY'ing the config, at the expense of workflow/status visibility - which isn't a worthwhile trade-off, IMO - https://gist.github.com/elithrar/4fa799c66b2c9932ac33f450f0787a58

The thing I'm mostly trying to solve for is versioning - I want it to be easy to run multiple versions (since we do so for backwards-compat!) and minimize human error when it comes to writing the CI config itself.

elithrar:
Going to merge and let co-exist with TravisCI for a while. Next steps would be to:

- [ ] Replace TravisCI entirely
- [ ] Script creation of a `.circleci/config.yml` for all gorilla/* repos based on their existing Travis CI config.

I'm also using this for another OSS project I'm about to release - to both test the library & build a container image - and the flexibility definitely pays off more there.

kisielk:
Would also be nice if the CircleCI script could somehow pick up new versions of Go into the build matrix. I'm not sure if that's possible or not, but that was always a maintenance headache with the travis config.

elithrar:
@kisielk - no, but that’d be great. A @circleci bot that can PR your `config.yml` with new versions would be nice though...

Part of the reason for wanting to write a script + template is so that we can code-gen creation of configs across repos and avoid skew. Now, to find the time for that...

elithrar:
For posterity: the script used to generate `config.yml` for every repo - https://gist.github.com/elithrar/3bf2e3bd60292e71d3b735cdab06cc78

PRs were still manual; those are next to automate - likely by traversing the repos under `$GOPATH/src/github.com/gorilla` and running a few `git` commands (branch, add, commit, push).

PR #493, opened 2019-06-29
Add documentation for using mux to serve a SPA

### LINKED ISSUE #464: Serve React Router
Go Version: 1.12.

Mux-Version: latest (not important)

Iam able to serve a react app which I created with "create-react-app", but how I serve an app which is created with "create-react-app" and has react router in it? Is this possible?

fharding1:
Fixes #464

**Summary of Changes**

1. Adds documentation to the README.md for using mux to serve API endpoints alongside a SPA.

**Testing**

I ran create-react-app, made a prod build, copied the build directory, and tested this out. Seems to work great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ func main() {
 
 Most of the time it makes sense to serve your SPA on a separate web server from your API,
 but sometimes it's desirable to serve them both from one place. It's possible to write a simple
-handler for serving your SPA (for use with React's BrowserRouter for example), and leverage
+handler for serving your SPA (for use with React Router's [BrowserRouter](https://reacttraining.com/react-router/web/api/BrowserRouter) for example), and leverage
 mux's powerful routing for your API endpoints.
 
 ```go
--- a/README.md
+++ b/README.md
@@ -234,30 +234,46 @@ import (
 	"github.com/gorilla/mux"
 )
 
+// spaHandler implements the http.Handler interface, so we can use it
+// to respond to HTTP requests. The path to the static directory and
+// path to the index file within that static directory are used to
+// serve the SPA in the given static directory.
 type spaHandler struct {
 	staticPath string
 	indexPath  string
 }
 
+// ServeHTTP inspects the URL path to locate a file within the static dir
+// on the SPA handler. If a file is found, it will be served. If not, the
+// file located at the index path on the SPA handler will be served. This
+// is suitable behavior for serving an SPA (single page application).
 func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+    // get the absolute path to prevent directory traversal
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -212,6 +213,77 @@ func main() {
...
+		return
+	} else if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	http.FileServer(http.Dir(h.staticPath)).ServeHTTP(w, r)
+}
+
+func main() {
```
elithrar: `func main()` LGTM

fharding1: Woah how did that happen. Swore I compiled this before PRing :thinking:

README.md
```diff
@@ -212,6 +213,77 @@ func main() {
...
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/gorilla/mux"
+)
+
+type spaHandler struct {
```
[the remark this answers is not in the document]
fharding1: Hmm. I can see how a `http.HandlerFunc` might be easier for a beginner than a type implementing `http.Handler`, however I think the overhead of the closure might outweigh that. Overall I think both are about as easy for a beginner to understand (also open to feedback of course :smile:).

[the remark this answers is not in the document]
elithrar: So, thinking on this: should we add a “ServeStatic” or similar handler as a built-in? This has come up a few times.

README.md
```diff
@@ -212,6 +213,77 @@ func main() {
...
+}
+
+func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	path, err := filepath.Abs(r.URL.Path)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	path = filepath.Join(h.staticPath, path)
```
elithrar: My general feedback here is that we (over) explain this function body.

fharding1:
I think implementing a built-in handler in `gorilla/mux` for the few people trying to use the wrong tool for the job (`gorilla/mux` to serve an SPA) wouldn't be worth adding more stuff to the API. Instead, I think it makes sense to leave this as docs so that people will at least stop opening issues asking how to serve an SPA with mux.

README.md
fharding1: Agreed, it should be more clear how this works.

fharding1:
Let me know if you think it still makes sense to add a built-in handler with the docs improvements and I'll implement that :+1:

elithrar:
Let's leave it as-is for now, @fharding1 - we can revisit later. `http.FileServer` continues to trip users up, so wary of that.

elithrar:
(Thanks for working on this, too!)

fharding1:
Np, thank you!

PR #501, opened 2019-07-20
Simplify code

muesli:
Use a single append call instead of a ranged for loop.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -111,10 +111,8 @@ func copyRouteConf(r routeConf) routeConf {
 		c.regexp.queries = append(c.regexp.queries, copyRouteRegexp(q))
 	}
 
-	c.matchers = make([]matcher, 0, len(r.matchers))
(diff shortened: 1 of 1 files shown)


mux.go
```diff
@@ -112,9 +112,7 @@ func copyRouteConf(r routeConf) routeConf {
 	}
 
 	c.matchers = make([]matcher, 0, len(r.matchers))
-	for _, m := range r.matchers {
-		c.matchers = append(c.matchers, m)
-	}
+	c.matchers = append(c.matchers, r.matchers...)
```
fharding1: IMO copy is a bit more clear here:

```
c.matchers = make([]matcher, len(r.matchers))
copy(c.matchers, r.matchers)
```

muesli: You're right, since we previously make sure there's enough space in the dst, we could just use copy instead. Will update the PR.

PR #517, opened 2019-09-02
Fixes path order dependent response when a nil handler is defined for a route

### LINKED ISSUE #515: [bug] Giving methodNotAllowed even when route is defined
**Describe the bug**
> When there is a route pattern with a different Method after the same pattern ( "OPTIONS /a/b
after GET /a/b") for which there is a middle-ware response, it gives 405. I guess it's related to #509 
…

**Versions**
> Go version: `go1.11.5 linux/amd64`
> package version: `9536e4053d763b54d935f1ce731a315cfb42b979`

…

**Steps to Reproduce**
> Define a route `/a/b` with method `GET`
> Define a route `/a/{a}` with method `OPTIONS`
> Define middleware which gives 200 for defined options routes
> It gives StatusMethodMismatch(405) instead of 200
> If the steps 1 and 2 are reversed it gives 200

…

**Expected behavior**
> The expected behaviour is to give 200

…

**Code Snippets**
>  mux_test.go
```


// testMiddleWare returns 200 on an OPTIONS request
func testOptionsMiddleWare(inner http.Handler) http.Handler {
		if r.Method == http.MethodOptions {
			w.WriteHeader(http.StatusOK)
			return
		}
		inner.ServeHTTP(w, r)
	})
}

// Passes if /a/{a} is defined first
func TestRouterOrder(t *testing.T) {
	handler := func(w http.ResponseWriter, r *http.Request) {}
	router := NewRouter()
	router.Path("/a/b").Handler(http.HandlerFunc(handler)).Methods(http.MethodGet)
	router.Path("/a/{a}").Handler(nil).Methods(http.MethodOptions)
	router.Use(MiddlewareFunc(testOptionsMiddleWare))

	w := NewRecorder()
	req := newRequest(http.MethodOptions, "/a/b")

	router.ServeHTTP(w, req)

	if w.Code != http.StatusOK {
		t.Fatalf("Expected status code 200 (got %d)", w.Code)
	}
}

```

…

**Possible Solution**

> `route.go` Line https://github.com/gorilla/mux/blob/master/route.go#L77
```
	diff --git a/route.go b/route.go
index 7343d78..1972c77 100644
--- a/route.go
+++ b/route.go
@@ -74,7 +74,7 @@ func (r *Route) Match(req *http.Request, match *RouteMatch) bool {
                return false
        }
 
-       if match.MatchErr == ErrMethodMismatch && r.handler != nil {
+       if match.MatchErr == ErrMethodMismatch && match.Handler == nil {
                // We found a route which matches request method, clear MatchErr
                match.MatchErr = nil
                // Then override the mis-matched handler
```

rkilingr:
Fixes #515

**Summary of Changes**

1. Checks route `handler != nil` condition only for setting the matched handler, not for clearing Method mismatch error
2. Adds test-cases to check order dependent response if request handler for the path is `nil`


mux_test.go
```diff
@@ -2803,6 +2803,48 @@ func TestSubrouterNotFound(t *testing.T) {
...
+	w := NewRecorder()
+	req := newRequest(http.MethodOptions, "/a/b")
+
+	router.ServeHTTP(w, req)
+
+	if w.Code != http.StatusOK {
+		t.Fatalf("Expected status code 200 (got %d)", w.Code)
+	}
+
+	reversedPathRouter := NewRouter()
```
elithrar: Break these into subtests - e.g.
```
t.Run("Test handler ordering", func(t *testing.T) { 
   // test 
})

t.Run("Test reversed handler ordering", func(t *testing.T) { 
   // test 
})
```

I would also add additional test cases:

1. Test ordering where handlers have the _same_ method & path, but a different handler (expected: the first handler added is the one executed)
2. Same path, different methods (no path variables)
3. Same as #2, but with middleware

mux_test.go
```diff
@@ -2803,6 +2803,48 @@ func TestSubrouterNotFound(t *testing.T) {
...
+	reversedPathRouter.Path("/a/{a}").Handler(http.HandlerFunc(handler)).Methods(http.MethodGet)
+	reversedPathRouter.Path("/a/b").Handler(nil).Methods(http.MethodOptions)
+	reversedPathRouter.Use(MiddlewareFunc(testOptionsMiddleWare))
```
elithrar: This should be before the paths are added.

mux_test.go
```diff
@@ -2803,6 +2803,48 @@ func TestSubrouterNotFound(t *testing.T) {
...
+
+	w = NewRecorder()
```
elithrar: Make it a subtest (as above) and call `newRequest` again to separate it.

route.go
```diff
@@ -74,11 +74,13 @@ func (r *Route) Match(req *http.Request, match *RouteMatch) bool {
 		return false
 	}
 
-	if match.MatchErr == ErrMethodMismatch && r.handler != nil {
+	if match.MatchErr == ErrMethodMismatch {
```
elithrar: Need more comments here, as this is a block that has changed multiple times. We should explain _why_ the code is this way - e.g.

```
// If a route matches, but the HTTP method does not, we do one of two (2) things:
// 1. Reset the match error if we find a matching method later.
// 2. Else, we override the matched handler in the event we have a possible fallback handler for that route.
//
// This is to prevent ...
```

(Feel free to suggest clearer/different documentation here).

elithrar (changes requested):

elithrar:
@rkilingr - did you still want to complete this?

pierreneter:
guys, please complete it

elithrar:
@rkilingr - just following up again if you're still keen on completing this?

rkilingr:
> @rkilingr - just following up again if you're still keen on completing this?

Apologies for the huge delay. Was not active last year, will be updating the PR

route.go
```diff
@@ -74,11 +74,13 @@ func (r *Route) Match(req *http.Request, match *RouteMatch) bool {
...
 		// We found a route which matches request method, clear MatchErr
 		match.MatchErr = nil
 		// Then override the mis-matched handler
-		match.Handler = r.handler
```
rkilingr, line 81: Will be removing this, as the use case this line was fixing in #317 was fixed subsequently with #422

route.go
rkilingr: Have added comments. Feel free to suggest modifications

mux_test.go
rkilingr: Have made the listed changes

amustaque97:
Hi @rkilingr - quick follow-up - are you interested in getting this merged? Will do a review again if existing comments are addressed.

jackgris (approved):

coreydaley (approved):

pcfreak30:
@coreydaley @jackgris @rkilingr this change really needs to get merged in tbh. im debugging my routing and the `if match.MatchErr != nil && r.handler != nil {` logic seems to do some really funky things when you have routes being tested before a valid one.

apoorvajagtap:
Hello @pcfreak30 Thanks for the work!
Corey is no longer active on the project, hence the pings went unaddressed. The maintainer details are available [here](https://github.com/orgs/gorilla/people).

I have just started looking into this issue. Would you be able to do a rebase to eliminate the conflicts?

pcfreak30:
> Hello @pcfreak30 Thanks for the work! Corey is no longer active on the project, hence the pings went unaddressed. The maintainer details are available [here](https://github.com/orgs/gorilla/people).
> 
> I have just started looking into this issue. Would you be able to do a rebase to eliminate the conflicts?

Um, this isn't my PR so im not sure what you mean.... Though I do want to see this bug fixed :P.

apoorvajagtap:
@pcfreak30 my bad. I mixed up two different communications. 
@rkilingr would you be able to rebase this?

PR #525, opened 2019-10-24
Remove/cleanup request context helpers

fharding1:
Fixes nothing

**Summary of Changes**

1. Removes context.go. Originally this was around so on Go < 1.7 we could use gorilla/context to "polyfill" context, but now that we've dropped support for that, I don't think it's necessary to have those helpers, plus it's nice to get rid of the (small) cognitive overhead of another file
2. Renames the request context functions from the pattern `setFoo` to `requestWithFoo`, since I think this makes it a bit more clear that they're returning a request with a new context wrapping the old one, not changing the request
3. Changes the request context functions to take concrete types instead of empty interfaces because there's no reason for them to take interfaces
4. Moves the single test in context_test.go to mux_test.go since that seems to be testing the muxer itself anyways, not really the context implementation or anything


elithrar:
Thanks!

PR #526, opened 2019-10-24
Fixed golint warnings

KosToZyB:
Fixed:
1. mux.go:366:5: error var SkipRouter should have name of the form ErrFoo
2. regexp.go:184:9: if block ends with a return statement, so drop this else and outdent its block

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -181,16 +181,16 @@ func (r *routeRegexp) Match(req *http.Request, match *RouteMatch) bool {
 			}
 		}
 		return r.regexp.MatchString(host)
-	} else {
-		if r.regexpType == regexpTypeQuery {
-			return r.matchQueryString(req)
-		}
-		path := req.URL.Path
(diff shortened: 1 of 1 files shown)


mux.go
```diff
@@ -361,9 +361,9 @@ func (r *Router) Walk(walkFn WalkFunc) error {
 	return r.walk(walkFn, []*Route{})
 }
 
-// SkipRouter is used as a return value from WalkFuncs to indicate that the
+// ErrSkipRouter is used as a return value from WalkFuncs to indicate that the
 // router that walk is about to descend down to should be skipped.
-var SkipRouter = errors.New("skip this router")
+var ErrSkipRouter = errors.New("skip this router")
```
elithrar: This is a breaking change. That golint complains is unfortunate, but this predates golint.

KosToZyB: Fixed

elithrar:
Thanks!

Issue #506, opened 2019-08-08
[docs] - 


RavenHursT:
Would be nice to have all the snippets in the documentation have complete import statements of all the packages referenced in each snippet code.

When a new Gopher comes along, trying to learn, and comes across a snippet like below.. It makes it really tough when snippets like this can't just be copy/pasted into things like, https://play.golang.org/ w/o first investigating and figuring out which packages need to be included.  There's nothing in the code snippet to _explicitly_ tell the Gopher where things like, `http`, `mux`, `httptest`, `fmt`, etc are coming from.

I know to a lot of seasoned folks, these things may seem "obvious" to them... But to new Gophers trying to learn how to use Golang, it's a big stumbling block...

![image](https://user-images.githubusercontent.com/496983/62718931-6fb34f80-b9cc-11e9-9740-d5b0595380c9.png)

elithrar:
This is totally fair - although note that external package support only
came to the Playground pretty recently. L

Are you interested in PR’ing these improvements?

RavenHursT:
Yeah.. I only used go playground as the example since it seems like the most widely used prototyping tool in Golangdom...  Personally, I've used Repl.it for years for a number of different server-side prototyping, and that's supported external packages for a while now..

Anyway.. yeah.. I could PR it.. if someone could maybe point me in the right direction... is it just `.md` file somewhere?

RavenHursT:
Like I said... baby Gopher here. :/

RavenHursT:
hmmm.. not seeing it here... https://github.com/gorilla/mux/blob/master/README.md#examples

RavenHursT:
Trying to find the file for this example specifically: https://godoc.org/github.com/gorilla/mux#CORSMethodMiddleware

elithrar:
Examples are typically in:

• files ending in _example.go
• the README.me
• doc.go

RavenHursT:
Ok... let me see if I can do it...

PR #532, opened 2019-11-08
Update first example

kunicmarko20:
**Summary of Changes**

1. Add a line that makes first example work

For me first example didn't work until I added `ListenAndServe`, if this is not needed feel free to close the pr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ func main() {
     r.HandleFunc("/products", ProductsHandler)
     r.HandleFunc("/articles", ArticlesHandler)
     http.Handle("/", r)
+    http.ListenAndServe(":8080", nil)
 }
 ```
 


fharding1:
Judging by the paragraph above:

```
Let's start registering a couple of URL paths and handlers:
```

It seems like this code block was only supposed to describe how to register the URL handlers and use gorilla/mux, not register them and start an HTTP server using the muxer, however, since it's including `func main()` it seems like it's supposed to be a full runnable example. In the latter case I think this change makes sense (and we should probably add import statements too), however maybe the `func main()` should just be excluded? This kind of reminds me of https://github.com/gorilla/mux/issues/506 . What do you think?

elithrar:
Fully runnable examples are good, and there is one further down too.

kunicmarko20:
Maybe make the first (or all examples) fully runnable since I didn't read the whole readme, my mistake. This change would probably remove confusion for new people.

elithrar:
Making them all fully runnable obscures the examples - since the (best
practice) code to make them runnable is often longer than the example.

kunicmarko20:
So. any thing you want me to update in this PR or should I just close it?

README.md
```diff
@@ -55,6 +55,7 @@ func main() {
     r.HandleFunc("/products", ProductsHandler)
     r.HandleFunc("/articles", ArticlesHandler)
     http.Handle("/", r)
```
elithrar: Replace both these lines with:

```go
log.Fatal(http.ListenAndServe(":8080", r))`
```

I would also add a comment above the handlers, along the lines of:

```go
// Each HandleFunc takes a path to match and a http.HandlerFunc to call on that match
```

README.md
```diff
@@ -55,6 +55,7 @@ func main() {
...
+    http.ListenAndServe(":8080", nil)
 }
```
elithrar: Replace both these lines with:

```go
log.Fatal(http.ListenAndServe(":8080", r))`
```

I would also add a comment above the handlers, along the lines of:

```go
// Each HandleFunc takes a path to match and a http.HandlerFunc to call on that match
```

elithrar:
There is also https://github.com/gorilla/mux#full-example in the README

PR #537, opened 2019-11-20
Add Cookies and CookiesRegexp to Router

icattlecoder:
This feature can be used for web applications when do rolling update deployment and a/b testing. It 's better than using headers regexp.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -306,6 +306,12 @@ func (r *Router) Headers(pairs ...string) *Route {
 	return r.NewRoute().Headers(pairs...)
 }
 
+// Cookies registers a new route with a matcher for request cookie values.
+// See Route.Cookies().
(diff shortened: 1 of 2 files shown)


fharding1:
Are there many other use cases for wanting to match routes based on cookies? A/B testing is potentially a fair one, but A/B testing isn't extremely widely used, and especially not A/B testing in this specific way (matching different routes based on cookies). Most of the time people will be using cookies for authentication, and likely won't care about matching different routes based on cookie values (at least while matching on raw strings).

Do you think this use case is common enough that it warrants adding a first party matcher, or do you think it makes more sense for people with this use case to use custom matchers?

elithrar:
I would suggest a custom matcher; A/B testing with only cookies doesn't
extend to logging, etc.

Since it can be done with middleware, I'd see that as making more sense
unless we see more demand for this.

PR #545, opened 2020-01-13
Added request logs

wesleimp:
This logs includes *status code*, *method*, *latency* and *path*


fharding1:
I'm not sure this is something mux should provide. Logging is something that almost *everyone* wants to do differently. Mux provides a way for people to easily hook up their own middleware for logging how they would like, which I think is enough. Additionally, a similar logging middleware is provided in gorilla/handlers for users already.

See:
* https://github.com/gorilla/mux/pull/46
* https://godoc.org/github.com/gorilla/handlers#LoggingHandler

wesleimp:
@fharding1 I got it. Well, I'm gonna close this PR and then I'll take a look at `gorilla/handlers` and other `gorilla` repositories and contribute for something.

What do you think about it?

PR #555, opened 2020-02-20
Add a flag to require explicit MethodNotAllowed handler

### LINKED ISSUE #554: [bug] The default MethodNotAllowedHandler is not spec compliant
[RFC 7231 Section 6.5.5](https://tools.ietf.org/html/rfc7231#section-6.5.5) 

States that 
> The origin server MUST generate an
>    Allow header field in a 405 response containing a list of the target
>    resource's currently supported methods.

Which is clearly not implemented right now
https://github.com/gorilla/mux/blob/75dcda0896e109a2a22c9315bca3bb21b87b2ba5/mux.go#L601-L603

Suggestion:
- Rather than assign a default MethodNotAllowedHandler just reuse the NotFoundHandler so it's both backward compatible and spec-compliant.

rocketspacer:
Fixes #554

**Summary of Changes**

1. Add a flag to handle MethodNotAllowed only when the MethodNotAllowedHandler is explicitly specified

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -51,6 +51,9 @@ type Router struct {
 	// Configurable Handler to be used when the request method does not match the route.
 	MethodNotAllowedHandler http.Handler
 
+	// If true, only handle MethodNotAllowed if the router MethodNotAllowedHandler is explicitly specified
+	RequireExplicitMethodNotAllowedHandler bool
+
 	// Routes to be matched, in order.
 	routes []*Route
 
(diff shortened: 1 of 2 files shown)


mux.go
```diff
@@ -51,6 +51,9 @@ type Router struct {
 	// Configurable Handler to be used when the request method does not match the route.
 	MethodNotAllowedHandler http.Handler
 
+	// If true, only handle MethodNotAllowed if the router MethodNotAllowedHandler is explicitly specified
+	RequireExplicitMethodNotAllowedHandler bool
```
elithrar, line 55: We can likely carry private state here as we know when the user has set the handler [above[

elithrar:
Sorry for the delay - can you expand on the docstring here @rocketspacer - ?

We should return the `Allow` header if a user specifies `.Methods` on a route, as we have that data, rather than grow the API or have divergent behaviour.

PR #579, opened 2020-07-10
fix: regression in vars extract for wildcard host

harshavardhana:
Continuing from PR #447 we have to add extra
check to ignore the port as well

### DIFF
--- a/old_test.go
+++ b/old_test.go
@@ -261,6 +261,18 @@ var hostMatcherTests = []hostMatcherTest{
 		result:  true,
 	},
 	{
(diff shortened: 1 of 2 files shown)


harshavardhana:
// cc @elithrar

elithrar:
Can you please add regression tests that capture this change? :)

harshavardhana:
@elithrar let me look at those - thanks for quickly looking at the PR.

harshavardhana:
would it be possible to make a release today if all goes well? @elithrar

elithrar:
I'll commit to having this reviewed before the end of the weekend.

harshavardhana:
@elithrar updated the PR with the tests, the same tests don't run with master branch.

harshavardhana:
Possible to tag a release @elithrar ?

harshavardhana:
> Possible to tag a release @elithrar ?

checking back to see if there can be a tagged release? @elithrar

PR #587, opened 2020-08-29
Stripe tail slash in PathPrefix

mrcyna:
As I was following the example and Test the API for the first time (BY THE WAY IT IS AWWWWSOME! & I LOVE IT)
I just had a problem in example and I figured out that maybe this would be misleading for other noob devs like me

Thanks so much

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ r.Host("{subdomain:[a-z]+}.example.com")
 There are several other matchers that can be added. To match path prefixes:
 
 ```go
-r.PathPrefix("/products/")
+r.PathPrefix("/products")
 ```
 
 ...or HTTP methods:


fharding1:
Could you elaborate on how the trailing slash was misleading?

mrcyna:
Hi

For example:
I write r.HandleFunc("/products/{key}", ProductHandler) and then add Prefix like this:
r.HandleFunc("/products/{key}", ProductHandler).PathPrefix("/foo/")

The route will be  "/foo//products/{key}" and I think double slashes somehow is misleading 

Thanks for your attention

elithrar:
It would make more sense to update the documentation for `PathPrefix` - a trailing slash has meaning here, and omitting it from the example may cause _more_ confusion.

https://pkg.go.dev/github.com/gorilla/mux?tab=doc#Route.PathPrefix

mrcyna:
You are right. I read the document again

Thanks for your attention

PR #589, opened 2020-08-31
allow skipping specified hosts from Host matcher

harshavardhana:
introduce SkipHosts(hosts ...string) at the router
level which can be used to allow certain hosts to be
ignored by the Host() matcher, this is some times
necessary for an application which registers handlers
for one domain to be at a completely different style
of operation

One example is S3 APIs when servicing them as path style
v/s bucket as part of the DNS style.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -82,6 +82,9 @@ type routeConf struct {
 	// will not redirect
 	skipClean bool
 
+	// If set to a host skips the this host from r.Host() matcher.
+	skipHosts map[string]struct{}
+
 	// Manager for the variables from host and path.
 	regexp routeRegexpGroup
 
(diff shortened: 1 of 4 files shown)


harshavardhana:
Is this something you would entertain? @elithrar

elithrar:
Another approach - that doesn't require API additions - would be to create _two_ host matchers, with one prioritized over the other, and then attach the same set of `Route` matchers.

This avoids the need to extend the API for this very specific use-case.

harshavardhana:
> Another approach - that doesn't require API additions - would be to create _two_ host matchers, with one prioritized over the other, and then attach the same set of `Route` matchers.

How would it work? I couldn't get it to work.

harshavardhana:
> Another approach - that doesn't require API additions - would be to create _two_ host matchers, with one prioritized over the other, and then attach the same set of `Route` matchers.

The reason is I need to skip a match, matchers by implementation match since Go doesn't support PCRE style negation regex its not possible to make it work to negate the matching behavior.  Unless you know some tricks which I couldn't find :cake:

harshavardhana:
actually was able to have similar style with 

```diff
+
        var routers []*mux.Router
        for _, domainName := range globalDomainNames {
-               routers = append(routers, apiRouter.Host("{bucket:.+}."+domainName).Subrouter())
+               routers = append(routers, apiRouter.MatcherFunc(func(r *http.Request, match *mux.RouteMatch) bool {
+                       host, _, _ := net.SplitHostPort(getHost(r))
+                       return host != minioReservedBucket+"."+domainName
+               }).Host("{bucket:.+}."+domainName).Subrouter())
```

PR #603, opened 2020-10-07
Update README.md

kennethtxytqw:
defer os_exit

**Summary of Changes**

1. Documentation update in `README.md`

I guess it makes more sense to phrase this as a question. If we do `os.exit(0)` when above there are deferred clean ups and e.g. `defer cancel()`. I believe that `os.exit(0)` skips all deferred calls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,7 +496,7 @@ func main() {
     // <-ctx.Done() if your application should wait for other services
     // to finalize based on context cancellation.
     log.Println("shutting down")
-    os.Exit(0)
+    defer os.Exit(0)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mrngm:
The [go documentation](https://golang.org/pkg/os/#Exit) agrees that deferred functions are not executed with `os.Exit(n)`. It would actually be better to remove `os.Exit(0)` here, because exit code 0 is default behaviour on termination.

kennethtxytqw:
I agree that `os.Exit(0)` is redundant here.

faith (approved):
Its for the sake of the status code however i a m pretty sure Go does it itself

PR #613, opened 2020-12-13
bugfix: set automatic redirection on routes without "/" on end of path

plankiton:
1. Always I to create a route on old version with "/user" as example, when I request "/user/" it returns a 404 error, to fix this I made the mux to create de route with "/" on end, and vice versa

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module github.com/gorilla/mux
+module github.com/plankiton/mux
 
 go 1.12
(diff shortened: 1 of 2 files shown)


Prithvipal:
Why the module name is changed?

https://github.com/gorilla/mux/pull/613/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R1

elithrar:
This would break a large number of existing applications who rely on prefix matching - implicitly creating routes is not the way to go.

PR #628, opened 2021-05-22
golint: fixed some golint warnings

sina-devel:
Fixed:

1. regexp_test.go:57:14: should omit 2nd value from range; this loop is equivalent to `for key := range ...`
2. regexp_test.go:82:14: should omit 2nd value from range; this loop is equivalent to `for key := range ...`


chutified (approved):
LGTM

codysnider:
LGTM also. This is being done on later PRs as well. May as well merge this in as an independent change to better isolate the later PRs.

Issue #354, opened 2018-03-13
Path incorrectly decoded during clean


cjellick:
**What version of Go are you running?** (Paste the output of `go version`)
go version go1.10 linux/amd64

**What version of gorilla/mux are you at?** (Paste the output of `git rev-parse HEAD` inside `$GOPATH/src/github.com/gorilla/mux`)
07ba1fd60e210dc5cc6f86ec25dc9ecc7c95c295

**Describe your problem** (and what you have tried so far)
The path cleaning performed in [Router.ServerHTTP()](https://github.com/gorilla/mux/blob/master/mux.go#L125) works solely on request.URL.Path which decodes all characters. So characters, particularly slashes, that were escaped are unescaped in the redirect location sent back to the user as part of the 301 redirect.

**Paste a minimal, runnable, reproduction of your issue below** (use backticks to format it)
```
package main

import (
	"fmt"
	"net/http"

	"github.com/gorilla/mux"
)

func handler(w http.ResponseWriter, req *http.Request) {
	fmt.Fprint(w, "hi!\n")
}
func main() {
	r := mux.NewRouter()
	r.HandleFunc("/hi", handler)

	server0 := &http.Server{
		Addr:    ":7777",
		Handler: r,
	}
	server0.ListenAndServe()

}
```

And then hit it with curl using a URL like this:
```
curl -i 'http://localhost:7777/asdf/..%2Fanotherpath'
HTTP/1.1 301 Moved Permanently
Location: /anotherpath
Date: Tue, 13 Mar 2018 16:08:34 GMT
Content-Length: 0
```
`%2F is and escaped `/` and should not be interpreted as a path segment delimiter.

In this case, the location of the redirect is `/anotherpath` but it should be `/..%2Fanotherpath`

elithrar:
* You can use `Route.SkipClean()` to avoid cleaning a specific route - https://godoc.org/github.com/gorilla/mux#Route.SkipClean - there is also a similar method on `Router`
* The path cleaning code mirrors `net/http` - https://github.com/gorilla/mux/blob/master/mux.go#L439-L456 - and isn't something we can change. Most users do want mux to clean such paths, which is why we provide an escape hatch for the few cases where you may not.

cjellick:
[...]

Issue #639, opened 2021-09-03
[bug] Wrong URL matching when url containing encoded slash (%2F)


gorshkov-leonid:
I found a lot of similar issues here but with some resolution results. But I think I ought to create this issue because think problem is still actual.

I use `/repositories/{id}/{branch}` and `/repositories/{id}/{branch}/files/structure`. When I use branch `feature/task1` (of cause, encoded as  `feature%2Ftask1`) then it is not working.
I am aware of:
* `router.UseEncodedPath()`
* `{branch:.+}` pattern

`router.UseEncodedPath()` disables auto decoding of path components. In this case, I have to decode it additionally. It is not handy.

`{branch:.+}` brokes behavior. For example `/repositories/{id}/{branch:.+}` matches `/repositories/123/feature%2Ftask1/files/structure`. So this is wrong way.

I think that `mux` works incorrectly. In my opinion URL matching should be done before decoding path components.

PR #641, opened 2021-09-07
Fix accidental double-escaping when Router.UseEncodedPath() is enabled

nvx:
Bug originally mentioned in https://github.com/gorilla/mux/issues/354#issuecomment-899377755

**Summary of Changes**

When Router.UseEncodedPath() is enabled (and SkipClean is not enabled), a request to `http://localhost/foo/../bar%25` causes a redirect to `http://localhost/bar%2525` as there is a bug where this code path results in double-encoding (the % from the %25 is itself encoded as %25, resulting in %2525).

The expected output for this should have been a redirect to `http://localhost/bar%25`. A unit test for this behaviour has been added which originally failed with the following output:

```
Expected redirect location to http://localhost/bar%25, found http://localhost/bar%2525
```

The remaining changes allows the new unit test to pass and does not break any existing unit tests.


nvx:
In an unrelated note, Router.UseEncodedPath() does not follow the same conventions as eg Router.SkipClean(bool) in that it is only possible to enable this function, and once called it is not possible to disable.

Due to the potential security vulnerabilities mentioned in #354 when UseEncodedPath is not enabled, arguably this should be the default, but then this would require a function to disable the behaviour if unwanted. Of interest setting UseEncodedPath by default does not break any existing unit tests.

nvx:
Something else that stood out when writing the unit tests for this is https://github.com/gorilla/mux/commit/d10d5466f2db2758bb616f627775692d2f69fe8a mentions some changes that were made to prevent the query string and # fragment from being stopped.

Oddly though looking at the commit to the Golang standard library it references https://github.com/golang/go/commit/b5842308892e0c4f9e772a42d5826f6f62f57be3 specifically mentions that it uses a relative URL to fix the issue, yet the behaviour of Gorilla Mux is to include an absolute URL in the Location header.

While the current fix does indeed include the query parameters, I'm not sure if there's other reasons for preferring a relative URL. Fragments are not sent to the server, so it's a bit harder to guess how browsers behave and if there's value in using a relative vs absolute redirect.

I'm happy to fix that up as well to make it match the standard library if that's what's desired, either as an additional commit in this PR, or as a separate PR after this is merged.

nvx:
I noticed #639 mentions that UseEncodedPath also affects behaviour elsewhere in decoding path components. Maybe the real answer is that UseEncodedPath shouldn't affect the behaviour of path cleaning and instead the path cleaning behaviour should always act as if it is enabled leaving UseEncodedPath to control the path matching behaviour only?

francescomari:
@nvx I think there is value in this PR, and I believe that they actually fix a genuine bug. But I'm new here, so take this with a generous grain of salt.

> I noticed #639 mentions that UseEncodedPath also affects behaviour elsewhere in decoding path components. Maybe the real answer is that UseEncodedPath shouldn't affect the behaviour of path cleaning and instead the path cleaning behaviour should always act as if it is enabled leaving UseEncodedPath to control the path matching behaviour only?

I'm not sure whether enabling UseEncodedPath by default right now is a good idea because, as you mention, it affects the behaviour around the unescaping of the captured variables. I worked for a fix related to this in #662. I think we could pull off changing the default behaviour of the router if we enable both UseEncodedPath and UnescapeVars (from #662) by default. I think that this wouldn't break the implementation of existing applications, even if they don't activate those flags in their own configuration.

> In an unrelated note, Router.UseEncodedPath() does not follow the same conventions as eg Router.SkipClean(bool) in that it is only possible to enable this function, and once called it is not possible to disable.

A similar conversation happened in #662, too. @elithrar offered some guidance about this and preferred to have UnescapeVars not accept any configurable flag in order to match the API of UseEncodedPath. I think this is a minor problem though, and the real focus should be defining what is the most secure default behaviour for the router, as long as we don't break existing code.

Some minor nitpicking: could you please rebase your PR onto the latest master? 91708ff fixed the build for the latest Go version, and that's the only thing that is making the checks fail for your PR.

nvx:
> @nvx I think there is value in this PR, and I believe that they actually fix a genuine bug. But I'm new here, so take this with a generous grain of salt.
> 
> > I noticed #639 mentions that UseEncodedPath also affects behaviour elsewhere in decoding path components. Maybe the real answer is that UseEncodedPath shouldn't affect the behaviour of path cleaning and instead the path cleaning behaviour should always act as if it is enabled leaving UseEncodedPath to control the path matching behaviour only?
> 
> I'm not sure whether enabling UseEncodedPath by default right now is a good idea because, as you mention, it affects the behaviour around the unescaping of the captured variables. I worked for a fix related to this in #662. I think we could pull off changing the default behaviour of the router if we enable both UseEncodedPath and UnescapeVars (from #662) by default. I think that this wouldn't break the implementation of existing applications, even if they don't activate those flags in their own configuration.
> 
> > In an unrelated note, Router.UseEncodedPath() does not follow the same conventions as eg Router.SkipClean(bool) in that it is only possible to enable this function, and once called it is not possible to disable.
> 
> A similar conversation happened in #662, too. @elithrar offered some guidance about this and preferred to have UnescapeVars not accept any configurable flag in order to match the API of UseEncodedPath. I think this is a minor problem though, and the real focus should be defining what is the most secure default behaviour for the router, as long as we don't break existing code.

Note that my changes do not break any existing tests. If we care about backwards compatibility we really need to first start by codifying what we consider valid uses of the library into some new unit tests, then work out how we can maintain compatibility and decide upon what edge cases we're comfortable with breaking in the name of "secure by default".

I daresay breaking a very small (perhaps nonexistent) number of users doing something sufficiently weird is going to be less bad overall than opening the majority of users up to some security vulnerabilities unless a few not very well understood options are enabled.

The way the params and escaping works currently is definitely a bit non-intuitive and I'm sure there's a better route we can take (perhaps matching on the unescaped path, pulling out the params, then unescaping the params after extraction for example) that might incur some minor breakages, but overall result in a better experience for all. The other option of course is the nuclear approach, cut a v2, deprecate v1 and tell everyone to move over with a warning that v1 has some undesirable security edge cases.

Needless to say these decisions are probably a bit larger than this PR, but I'd be happy to throw some commits at it if we can decide on which way forward we want to take.

> Some minor nitpicking: could you please rebase your PR onto the latest master? [91708ff](https://github.com/gorilla/mux/commit/91708ff8e35bafc8612f690a25f5dd0be6f16864) fixed the build for the latest Go version, and that's the only thing that is making the checks fail for your PR.

Done

nvx:
this is still relevant

amustaque97:
> this is still relevant

Hey @nvx, sorry for the delay. Are you still following up with the PR? I will be reviewing the PR by this weekend.

nvx:
> > this is still relevant
> 
> Hey @nvx, sorry for the delay. Are you still following up with the PR? I will be reviewing the PR by this weekend. 

Yup. Will be good to see some movement on this.

jarri-abidi (approved):

PR #647, opened 2021-10-15
Create helper method for creating middlewares from a closure

marianososto:
**Context**:
It's not possible to create a middleware just by declaring a closure or anonymous function. So I've create a helper method for creating a Middleware just by declaring the anonymous function.

Fixes #
- This PR doesn't fix any bug nor issue, it's just new functionality which makes it more comfortable to create a middleware.

**Summary of Changes**

1.Create new helper method for creating Middleware from a closure or anonymous function.
2. Unit test was added to test the new functionality.

Example:

```
NewMiddleware(func(http.ResponseWriter, *http.Request) {
 // TODO: here write the code of your middleware
})
```

Another option is to declare the function in a separate variable

```
myMiddleware := func(http.ResponseWriter, *http.Request) {
// TODO: do some stuff
}
NewMiddleware(myMiddleware)
```

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -37,6 +37,7 @@ func NewMiddleware(mw func(http.ResponseWriter, *http.Request)) MiddlewareFunc {
 type genericMW struct {
 	process func(http.ResponseWriter, *http.Request)
 }
+
 func (instance *genericMW) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	instance.process(w, req)
 }
--- a/middleware.go
+++ b/middleware.go
@@ -20,6 +20,27 @@ func (mw MiddlewareFunc) Middleware(handler http.Handler) http.Handler {
 	return mw(handler)
 }
 
+// NewMiddleware helper method which allows creating a middleware from a function which receives the response and request.
+// The typical case would be passing a closure
+func NewMiddleware(mw func(http.ResponseWriter, *http.Request)) MiddlewareFunc {
+	return func(handler http.Handler) http.Handler {
+		return &genericMW{
+			process: func(w http.ResponseWriter, r *http.Request) {
+				mw(w, r)
+				handler.ServeHTTP(w, r)
+			},
+		}
+	}
+}
+
+// genericMW is for creating a middleware from a closure, it's used in the public method NewMiddleware
+type genericMW struct {
+	process func(http.ResponseWriter, *http.Request)
+}
+func (instance *genericMW) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	instance.process(w, req)
+}
+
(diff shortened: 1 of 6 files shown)


codysnider:
LGTM, but #628 should be merged first and conflicts will need to be resolved.

mux.go
```diff
@@ -363,9 +363,9 @@ func (r *Router) Walk(walkFn WalkFunc) error {
 	return r.walk(walkFn, []*Route{})
 }
 
-// SkipRouter is used as a return value from WalkFuncs to indicate that the
+// ErrSkipRouter is used as a return value from WalkFuncs to indicate that the
 // router that walk is about to descend down to should be skipped.
-var SkipRouter = errors.New("skip this router")
+var ErrSkipRouter = errors.New("skip this router")
```
francescomari, line 368: You can't rename `SkipRouter` to `ErrSkipRouter` without breaking existing user code. This change is not backwards compatible. You should revert this rename.

middleware.go
```diff
@@ -20,6 +20,28 @@ func (mw MiddlewareFunc) Middleware(handler http.Handler) http.Handler {
 	return mw(handler)
 }
 
+// NewMiddleware helper method which allows creating a middleware from a function which receives the response and request.
+// The typical case would be passing a closure
+func NewMiddleware(mw func(http.ResponseWriter, *http.Request)) MiddlewareFunc {
```
francescomari, line 25: The purpose of `NewMiddleware` is not clear to me. IIUC, `NewMiddleware` takes an HTTP handler A as input and produces a `MiddlewareFunc` as output. When the `MiddlewareFunc` is invoked to wrap an HTTP handler B, it produces a new HTTP handler C that invokes A and B in sequence.

I personally fail to see how this can be useful. Can you please explain in more detail the problem that you are trying to solve?

middleware.go
```diff
@@ -20,6 +20,28 @@ func (mw MiddlewareFunc) Middleware(handler http.Handler) http.Handler {
...
+
+// genericMW is for creating a middleware from a closure, it's used in the public method NewMiddleware
+type genericMW struct {
+	process func(http.ResponseWriter, *http.Request)
+}
+
+func (instance *genericMW) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	instance.process(w, req)
+}
+
```
francescomari, line 44: IMO a new type to support the implementation of `NewMiddleware` is a bit overkill. You could implement `NewMiddleware` with closures all the way down:

```go
func NewMiddleware(mw func(http.ResponseWriter, *http.Request)) MiddlewareFunc {
	return func(handler http.Handler) http.Handler {
			mw(w, r)
			handler.ServeHTTP(w, r)
		})
	}
}
```

francescomari (changes requested):

PR #652, opened 2021-11-07
Set the Allow header when ErrMethodMismatch is set

gduchene:
👋

As described in #554, the current default behavior of mux isn’t compliant with [RFC 7231§6.5.5][RFC7231] when 405 Method Not Allowed is returned.

This PR makes it so that `ErrMethodMismatch` becomes a struct containing the slice of methods that would be allowed by the matcher, which then makes it possible for `Router.ServeHTTP` to construct a handler that will set the `Allow` header, which solves the issue.

I also considered adding a field in `Route` that would have that slice as well (as discussed in #555), but it felt neater to set an error with the information needed. It also allows user-defined matchers to return that error, which mux would then handle correctly. Happy to reconsider, of course!

Fixes https://github.com/gorilla/mux/issues/554.

[RFC7231]: https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.5

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"path"
 	"regexp"
+	"strings"
 )
 
 var (
@@ -202,7 +203,7 @@ func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	if handler == nil && match.MatchErr == ErrMethodMismatch {
-		handler = methodNotAllowedHandler()
+		handler = methodNotAllowedHandler(match.AllowedMethods)
 	}
 
 	if handler == nil {
@@ -417,6 +418,10 @@ type RouteMatch struct {
 	// It is set to ErrMethodMismatch if there is a mismatch in
 	// the request method and route method
 	MatchErr error
+
+	// AllowedMethods contains the list of methods allowed by a route when
+	// MatchErr is set to ErrMethodMismatch.
+	AllowedMethods []string
 }
 
 type contextKey int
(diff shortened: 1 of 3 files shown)


mux.go
```diff
@@ -11,15 +11,22 @@ import (
...
+// ErrMethodMismatch is returned when the method in the request does not
+// match the method defined against the route.
+type ErrMethodMismatch struct {
+	// The methods that the matcher would have allowed.
+	AllowedMethods []string
+}
+
+func (ErrMethodMismatch) Error() string {
+	return "method is not allowed"
+}
```
francescomari, line 26: Changing `ErrMethodMismatch` this way is not backwards compatible. Existing user code checking for `err == ErrMethodMismatch` will break after this change because `ErrMethodMismatch` is not a value anymore.

francescomari (changes requested):
I agree with the fix, but I disagree with the implementation. This PR changes the API of the package in a backwards incompatible way.

We need to find another way to pass the list of allowed HTTP methods from a `Route` to its `Router`. What about an additional field `AllowedMethods` in `RouteMatch`, to be populated only if `MatchErr == ErrMethodMismatch`?

mux.go
gduchene: You’re right, and I agree with your suggestion. Will change the PR accordingly.

gduchene:
OK, this should be enough. I’ve also removed the other commits as they are no longer in scope for this PR.

gduchene:
Hey @francescomari, quick ping on this—I can’t request another review so maybe this PR fell through the cracks! 🙂

francescomari:
@gduchene sorry for the delay, this definitely fell through the cracks. As much as I had fun reviewing your PR, I have to point out that I'm not a maintainer for this project. I started reviewing and opening PRs here to help the only maintainer left (see #659). I'm not saying this to discourage further contributions, but to set the right expectations. It was a pleasure reviewing your PR (and I've learned something in the process, too), but only a maintainer can push your change forward from now on.

gduchene:
Ahh, I didn’t realize that. Thanks for the review anyway! @elithrar mind taking a look when you have the time?

amustaque97:
> Ahh, I didn’t realize that. Thanks for the review anyway! @elithrar mind taking a look when you have the time?

@elithrar a gentle remainder.

PR #658, opened 2021-12-08
Enhancement for shutdown example

paul-nelson-baker:
**Summary of Changes**

The current example for graceful-shutdown is already very complete, but
can be improved for situations where the server cannot start correctly
from the beginning.

There was no executable code changed, and there are no executable examples relevant to this. If you would like me to create one to include I'm more than happy to accommodate that inclusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -472,31 +472,40 @@ func main() {
     }
 
     // Run our server in a goroutine so that it doesn't block.
+    serverErrorChannel := make(chan error, 1)
+    defer close(serverErrorChannel)
     go func() {
-        if err := srv.ListenAndServe(); err != nil {
-            log.Println(err)
-        }
+        log.Printf("Running server: %s\n", server.Addr)
+        serverErrorChannel <- server.ListenAndServe()
     }()
 
-    c := make(chan os.Signal, 1)
+    osSignalChannel := make(chan os.Signal, 1)
+    defer close(osSignalChannel)
     // We'll accept graceful shutdowns when quit via SIGINT (Ctrl+C)
     // SIGKILL, SIGQUIT or SIGTERM (Ctrl+/) will not be caught.
-    signal.Notify(c, os.Interrupt)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -472,31 +472,40 @@ func main() {
...
+    select {
+    case capturedSignal := <-osSignalChannel:
+        log.Printf("Signal caught: %s", capturedSignal)
+        // Create a deadline to wait for.
+        ctx, cancel := context.WithTimeout(context.Background(), wait)
+        defer cancel()
+        // Doesn't block if no connections, but will otherwise wait
+        // until the timeout deadline.
+        server.Shutdown(ctx)
+        <-serverErrorChannel
```
francescomari, line 498: I think it's not necessary to wait for `serverErrorChannel` to be closed. According to the documentation,

> When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return ErrServerClosed. Make sure the program doesn't exit and waits instead for Shutdown to return.

Waiting until `server.Shutdown()` returns should be enough.

README.md
```diff
@@ -472,31 +472,40 @@ func main() {
...
+    // Block until we receive from one of our channels
+    select {
+    case capturedSignal := <-osSignalChannel:
+        log.Printf("Signal caught: %s", capturedSignal)
+        // Create a deadline to wait for.
+        ctx, cancel := context.WithTimeout(context.Background(), wait)
+        defer cancel()
+        // Doesn't block if no connections, but will otherwise wait
+        // until the timeout deadline.
+        server.Shutdown(ctx)
```
francescomari, line 497: It would be nice to handle the error returned by `server.Shutdown()`. Sometimes the error can be useful, since the method might return any errors occurred while shutting down the Listeners.

francescomari (changes requested):
It would be nice if you could rebase this PR on the latest master. The change at 91708ff fixed the issue that is making the build fail for your PR, too.

PR #661, opened 2021-12-15
add golintci linter

sukhjit:
**Summary of Changes**

1. Added golangci-lint linter
2. Makefile for "lint" and "test"
3. Updated code as per issues raised by linter


Makefile
```diff
@@ -0,0 +1,13 @@
+GOBIN=$(shell pwd)/bin
```
Volatus, line 1: Isn't it a bit excessive to add an entire Makefile just for this? I think instructions could just be part of a `CONTRIBUTING.md` or the `README.md`.

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+install-lint:
+	@if [ ! -f "$(GOBIN)/golangci-lint" ]; then \
+		curl -sSfL -q https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(GOBIN) v1.42.0; \
```
Volatus, line 5: I'm personally not a fan of just running shell scripts to install tooling like this, it makes more sense to install it in a more package manager friend way. (Maybe just point people to the `golangci-lint` project instead where they have platform specific instructions on how to properly install it.

route.go
```diff
@@ -64,7 +64,6 @@ func (r *Route) Match(req *http.Request, match *RouteMatch) bool {
 				match.MatchErr = nil
 			}
 
-			matchErr = nil
```
Volatus, line 67: This change feels a bit outside the scope of this particular PR.

Makefile
sukhjit: Putting this in makefile makes it easier for a user to run the command. In this case a developer can simply check for linting using: `make lint`. Most of the Golang projects use `Makefile` to make it easier to run the commands including `test` and `build`.

Some examples:
* https://github.com/uber-go/zap/blob/master/Makefile
* https://github.com/hashicorp/terraform/blob/main/Makefile

route.go
sukhjit: Golint pointed out that this is a `deadcode`, that's why I've removed. I can revert this part and mark it with `// nolint` for now.

Makefile
sukhjit: I reason I used it this way to install was that it covers `Linux`, `Mac` and to some extent `Windows` (using Git Bash or WSL). I wanted to install it using `go get`, but that method is not recommended.
```
Note: such go get installation aren't guaranteed to work. We recommend using binary installation.
```

Volatus: Understood, could you add a conditional so that on macOS it strictly uses `brew` and not the install script? (`brew install golangci-lint`)

sukhjit: The way `make lint` currently works is that it installs `golangci-lint` inside `bin` directory of this repo. If I change it to use `brew` to install the linter, it will install `golangci-lint` on developer's machine. I don't think it a good strategy to install software on developer's machine, that's only relevant to this repo.

Makefile
```diff
@@ -0,0 +1,13 @@
...
+	fi
+
+lint: install-lint
+	@echo "golangci-lint run ../..."
```
RaboliotLeGris, line 9:
```suggestion
	@echo "golangci-lint run ./..."
```

sukhjit: Fixed

Makefile
bluebrown: what about tools.go?   https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

coreydaley:
This has been added during the transition

PR #662, opened 2021-12-17
Add a flag to unescape variable values

### LINKED ISSUE #639: [bug] Wrong URL matching when url containing encoded slash (%2F)

francescomari:
Fixes #639

**Summary of Changes**

1. Add a new router option configurabile via `UnescapeVars`.
2. When `UnescapeVars` is active, variable values are automatically unescaped when a match is found.
3. When `UnescapeVars` is active, variable values are automatically escaped before generating the route's URL.

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -84,6 +84,8 @@ type routeConf struct {
 	// will not redirect
 	skipClean bool
 
+	unescapeVars bool
+
 	// Manager for the variables from host and path.
 	regexp routeRegexpGroup
 
@@ -257,6 +259,14 @@ func (r *Router) StrictSlash(value bool) *Router {
 // become /fetch/http/xkcd.com/534
 func (r *Router) SkipClean(value bool) *Router {
 	r.skipClean = value
+	return r
+}
+
+// UnescapeVars tells the router to escape the route variables before invoking
+// the handler. This is useful when used together with UseEncodedPath to avoid
+// un-escaping the variable values in the handler.
+func (r *Router) UnescapeVars(value bool) *Router {
+	r.unescapeVars = value
 	return r
 }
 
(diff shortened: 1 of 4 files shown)


francescomari:
@elithrar I tried to fix #639 with as few changes as possible, but I figured that it would have been difficult to maintain backwards compatibility without adding another option to the `Router`. If you think this is not the right way to go, please provide guidance on how to proceed (or whether #639 should be fixed at all).

mux.go
```diff
@@ -260,6 +262,14 @@ func (r *Router) SkipClean(value bool) *Router {
 	return r
 }
 
+// UnescapeVars tells the router to escape the route variables before invoking
+// the handler. This is useful when used together with UseEncodedPath to avoid
```
elithrar, line 266: Can you add an example to the godoc here too, so that readers have a practical example to wrap their head around?

elithrar (changes requested):

francescomari:
@elithrar I have a question re. the consistency of the API. `UseEncodedPath`  receives no arguments, but `UnescapeVars` does receive a boolean. Wouldn't look more consistent if `UnescapeVars` received no arguments, too? If called, it would just activate the flag, like `UseEncodedPath` does.

mux.go
francescomari: Done at 7885b61.

elithrar:
> Wouldn't look more consistent if UnescapeVars received no arguments, too? If called, it would just activate the flag, like UseEncodedPath does.

Yes, unlabeled (Go doesn't have labelled args), single Boolean arguments are a bit of an anti-pattern.

francescomari:
I've removed the flag. Now `UnescapeVars` just set the flag as `true` under the hood.

PR #598, opened 2020-09-20
pattern aliases

HiveTraum:
This feature introduces ability to register aliases for some often used regular expressions. 

For example path as this:
`/{category:[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}}/{product:[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}}`
can be transformed into this
`/{category:uuid}/{product:uuid}`

You need just use register it
```
router := NewRouter()
router.RegisterPattern("uuid", "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}")
router.Path("/{category:uuid}")
router.Path("/{product:uuid}")
router.Path("/{order:uuid}")
router.Path("/{user:uuid}")
```

1. New method `RegisterPattern` on `Route` struct
2. New method `RegisterPattern` on `Router` struct

PR #675, opened 2022-05-15
pattern aliases

HiveTraum:
✋Reopened #598 

This feature introduces ability to register aliases for some often used regular expressions. 

For example path as this:
`/{category:[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}}/{product:[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}}`
can be transformed into this
`/{category:uuid}/{product:uuid}`

You need just use register it
```
router := NewRouter()
router.RegisterPattern("uuid", "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}")
router.Path("/{category:uuid}")
router.Path("/{product:uuid}")
router.Path("/{order:uuid}")
router.Path("/{user:uuid}")
```

1. New method `RegisterPattern` on `Route` struct
2. New method `RegisterPattern` on `Router` struct


example_router_test.go
```diff
@@ -0,0 +1,25 @@
+package mux_test
```
amustaque97, line 1: We should move this test case in the file: `example_route_test.go`

mux.go
```diff
@@ -124,6 +126,14 @@ func copyRouteRegexp(r *routeRegexp) *routeRegexp {
 	return &c
 }
 
+func (r *Router) RegisterPattern(alias string, pattern string) *Router {
```
amustaque97, line 129: Can we add comments here?

mux.go
```diff
@@ -94,6 +94,8 @@ type routeConf struct {
 	buildScheme string
 
 	buildVarsFunc BuildVarsFunc
+
```
amustaque97, line 97: Same, can we add comment here as well?

mux_test.go
```diff
@@ -216,6 +216,16 @@ func TestHost(t *testing.T) {
...
 		},
+		{
+			title:        "Host route with alias patterns",
+			route:        new(Route).RegisterPattern("version", "[a-z]{3}").Host("{v-1:version}.{v-2:version}.{v-3:version}"),
+			request:      newRequest("GET", "http://aaa.bbb.ccc/111/222/333"),
+			vars:         map[string]string{"v-1": "aaa", "v-2": "bbb", "v-3": "ccc"},
+			host:         "aaa.bbb.ccc",
+			path:         "",
+			hostTemplate: `{v-1:version}.{v-2:version}.{v-3:version}`,
+			shouldMatch:  true,
```
amustaque97, line 227: Can we add negative test case where `shouldMatch` is false because of mismatch.

mux_test.go
```diff
@@ -449,6 +459,36 @@ func TestPath(t *testing.T) {
...
+			request:      newRequest("GET", "http://localhost/dce51145-5cc3-4b54-bfb0-7bdb64a67e4d/a385ddcb-278e-4234-93dd-4d7b0fcb95c1"),
+			vars:         map[string]string{"category": "dce51145-5cc3-4b54-bfb0-7bdb64a67e4d", "product": "a385ddcb-278e-4234-93dd-4d7b0fcb95c1"},
+			host:         "",
+			path:         "/dce51145-5cc3-4b54-bfb0-7bdb64a67e4d/a385ddcb-278e-4234-93dd-4d7b0fcb95c1",
+			pathTemplate: `/{category:uuid}/{product:uuid}`,
+			shouldMatch:  true,
+		},
+		{
+			title:        "Path route with regexp alias patterns passed through router",
+			route:        NewRouter().RegisterPattern("digits", "[0-9]+").Path("/{id:digits}"),
```
amustaque97, line 484: Both the test cases look similar to me.
1. First one:
```javascript
{
	title:        "Path route with regexp alias patterns",
	route:        new(Route).RegisterPattern("digits", "[0-9]+").Path("/{id:digits}"),
	request:      newRequest("GET", "http://localhost/1"),
	vars:         map[string]string{"id": "1"},
	host:         "",
	path:         "/1",
	pathTemplate: `/{id:digits}`,
	shouldMatch:  true,
}
```

2. Second:
```javascript
{
	title:        "Path route with regexp alias patterns passed through router",
	route:        NewRouter().RegisterPattern("digits", "[0-9]+").Path("/{id:digits}"),
	request:      newRequest("GET", "http://localhost/1"),
	vars:         map[string]string{"id": "1"},
	host:         "",
	path:         "/1",
	pathTemplate: `/{id:digits}`,
	shouldMatch:  true,
}
```

We should add at least one negative test case. What do you think?

amustaque97:
Thank you @HiveTraum for bringing this up again. I have added a couple of comments. I need some more time to review it completely. In the meantime, I Would love to hear from you.

mux.go
HiveTraum: Added comments

mux.go
HiveTraum: Added comments

mux_test.go
HiveTraum: Added negative test case

mux_test.go
HiveTraum: Yeah, seems like this tests was very similar, deleted second test and added negative test case

example_router_test.go
HiveTraum: modev example

amustaque97:
@elithrar , I reviewed the PR and changes looks good to me and ready to ship 🚀 

I would request you to please take a final look at this and if all good. Can we merge it?

HiveTraum:
Yes, we can merge this

PR #676, opened 2022-05-18
Add GetVarNames()

eh-steve:
**Summary of Changes**

1. Added `r.GetVarNames()` function to list all vars a route might need in order to call `r.URL()`


amustaque97:
I will review PR by this weekend.

eh-steve:
The reason for this is where routes are built up programmatically from a configuration, not hard coded by a developer. In this case, the application needs to know the registered vars to be able to ensure they are available to be injected (or retrieve them if necessary)

This may not be a common use case but we're using mux in this way.

eh-steve:
I'm not sure I follow what you mean about choices - right now, there's no way of knowing which vars a route captures (on the basis that the caller might not be the one who built the route), and this adds a fairly non-intrusive way to expose that information.

An alternative choice for would be to force users to make another struct which embeds the mux route and keeps track of all var names by duplicating/wrapping a lot of functionality?

If we wanna freeze the API of this library that's fine, you can close this PR and I can use the fork?

amustaque97:
> If we wanna freeze the API of this library that's fine

Not want to make any premature decision here. Could you please share the minimal code(building of URL programmatically) where the current PR method `GetVarNames` comes into play? Later we can discuss whether we should merge it or not.

eh-steve:
Sure, so:

Imagine an application that makes HTTP requests based on variables coming from an inbound messages (e.g. from a kafka topic). It could take a set of YAML configurations like:

```yaml
endpoint:
  host_pattern: "www.{domain}.com"
  path_pattern: "/some/prefix/{group}/{item_id:[0-9]+}"
  method: "POST"
  queries:
    blah: "{some_data1}"
    something: "{some_data2}"
```
and builds up mux routes programatically using these configured endpoints.

Then when a payload comes in from an inbound message like:
```json
{
  "domain": "example",
  "group": "my_group",
  "item_id": 75,
  "some_data1": "123",
  "some_data2": "456",
  "other_irrelevant_data": {
    "massive_and_expensive_to_stringify": "..."
  }
}
```
It would be able to extract keys from the JSON payload (parsed as a `map[string]interface{}`) and use them as URL vars. Mux vars values need to be strings, so any non-string values would need to be `fmt.Sprint()`'ed (e.g. `item_id` in this example).

If your JSON payload contains any additional fields which aren't needed as route vars (e.g. `other_irrelevant_data` in this example, which contains the field `massive_and_expensive_to_stringify`), we would still need to stringify them all (which involves a needless allocation) and pass them into `route.URL(pairs...)`, where they would be ignored anyway.

By exposing `GetVarNames()`, it allows us to only pass variables which we know are required by the route, and to error early if those variables are not available, or don't adhere to a schema etc.

This is a simplified example compared to how we're actually using it, but it illustrates the point.

jrs526 (approved):
This seems like a harmless change that could benefit some users.  I think in this case, the change would have to be identified as a maintenance risk to justify NOT accepting the PR.

route.go
```diff
@@ -718,6 +718,25 @@ func (r *Route) GetHostTemplate() (string, error) {
 	return r.regexp.host.template, nil
 }
 
+// GetVarNames returns the names of all variables added by regexp matchers
+// These can be used to know which route variables should be passed into r.URL()
```
amustaque97, line 722: Hi @eh-steve, can we add an example in the docs? I agree you already explain the use case briefly in the PR comment. Others who will go through the package documentation should understand its use case properly.

```go
func main() {

	r := mux.NewRouter()
	r.Host("{domain}").
		Path("/{group}/{item_id}").
		Queries("some_data1", "{some_data1}").
		Queries("some_data2", "{some_data2}").
		HandlerFunc(Func).
		Name("article")
	fmt.Println(r.Get("article").GetVarNames())
	http.ListenAndServe(":8000", r)

}
```` 

`GetVarNames` will print `[domain group item_id some_data1 some_data2] <nil>`

Comment (amustaque97) [reply route.go:722]: Can we add an example in the `README.md` file and add test cases to validate the implementation.

Comment (eh-steve) [reply route.go:722]: I've added some, not sure how useful they are, but lemme know?

**Comment by eh-steve** [comment]
> To remove curly braces {} and slash / writing half of the function identical to newRouteRegexp: https://github.com/gorilla/mux/blob/master/regexp.go#L41 doesn't sounds like a good idea 

That's how I first started (on the calling side) before I realised it was a bad idea!

File: mux_test.go [hunk 2]
```diff
@@ -2828,6 +2828,40 @@ func TestContextMiddleware(t *testing.T) {
...
+func TestGetVarNames(t *testing.T) {
+	r := NewRouter()
+
+	route := r.Host("{domain}").
+		Path("/{group}/{item_id}").
+		Queries("some_data1", "{some_data1}").
+		Queries("some_data2_and_3", "{some_data2}.{some_data3}")
+
+	// Order of vars in the slice is not guaranteed, so just check for existence
+	expected := map[string]bool{
```
amustaque97, line 2840: Change expected to this:

```go
expected := []string{"domain", "group", "item_id", "some_data1", "some_data2", "some_data3"}
```

mux_test.go
```diff
@@ -2828,6 +2828,40 @@ func TestContextMiddleware(t *testing.T) {
...
+	varNames, err := route.GetVarNames()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(varNames) != len(expected) {
+		t.Fatalf("expected %d names, got %d", len(expected), len(varNames))
+	}
+
+	for _, varName := range varNames {
```
amustaque97, line 2858: You can use `reflect.DeepEqual` to check if `varNames`  and `expected` slice are equal or not.

route.go
amustaque97: Added a review comment for the test case.

mux_test.go
eh-steve: This is deliberate, since I don't want to enforce any guarantees about slice ordering in the test.

amustaque97 (commented):
One last comment I have added, and then I will approve the PR. 

@eh-steve Good work! Thanks for quickly tackling this.

amustaque97 (approved):
Hi @elithrar, please merge this PR. 👍🏻 

**Notice that there is no CI run for this PR. We need to fix CI.**

jackgris (approved):

coreydaley:
@apoorvajagtap Can you please take a look at this an approve if it looks good?

apoorvajagtap (approved):

PR #678, opened 2022-06-01
Fix  `Single Page Application` example in `README.md` file

### LINKED ISSUE #588: [question] Gorilla/Mux SPA not working
**Problem**

So I'm not really sure if this is a bug or anything, and this is also my first bug report so I apologize for my lack of knowledge on correctly reporting a bug.

I was trying to create a SPA using gorilla/mux and kept running into issues with displaying my webpage. I knew my code was correct but it had something to do with the file paths inside the SPA handler itself because when I was running my code I was getting a StatusInternalServerError. All my file paths were correct so it had to be something with the ServeHTTP function itself. I later found it to be an issue with the path as shown below.

**Versions**

go version go1.13.4 windows/amd64

github.com/gorilla/mux v1.7.4
…

**main.go**
```
type spaHandler struct {
	staticPath string
	indexPath  string
}

func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	path, err := filepath.Abs(r.URL.Path)
	if err != nil {
		http.Error(w, err.Error(), http.StatusBadRequest)
		return
	}

	path = filepath.Join(h.staticPath, path) //this is the part of the code I changed to fix my issue

	_, err = os.Stat(path)
	if os.IsNotExist(err) {
		http.ServeFile(w, r, filepath.Join(h.staticPath, h.indexPath))
		return
	} else if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}

	http.FileServer(http.Dir(h.staticPath)).ServeHTTP(w, r)
}

func main() {
	router := mux.NewRouter()
	spa := spaHandler{staticPath: "./admin/build", indexPath: "index.html"}
	router.PathPrefix("/").Handler(spa)
	log.Fatal(http.ListenAndServe(":8080", router))
}
```

Code was changed to below, and started to work perfectly fine.
```
path = filepath.Join(h.staticPath, r.URL.Path) 
```
…

I don't know whether or not to report this as a fix to the actual public code or not. Still new to contributing.

amustaque97:
Fixes #588

**Summary of Changes**

1. Add test case to validate proposed fix (both negative and positive test case).
2. Update `README.md` file.

PS: If you want to verify how I'm able to reproduce the issue and tried my fix. Here is a link of my local PR where I have configured `Github Action` to execute same test cases on all the platforms i.e. `ubuntu`, `macos` and `windows`. 

PR link: https://github.com/amustaque97/mux/pull/1


amustaque97:
@elithrar can you please review the PR? Looking forward to hearing from you. 
Thank you

README.md
```diff
@@ -251,18 +251,16 @@ type spaHandler struct {
 // file located at the index path on the SPA handler will be served. This
 // is suitable behavior for serving an SPA (single page application).
 func (h spaHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-    // get the absolute path to prevent directory traversal
-	path, err := filepath.Abs(r.URL.Path)
+	var err error
```
casserlyprogramming, line 254: how would err ever get set here? we initialise it but never git it a value unless I'm misunderstanding?

elithrar: To add: err will always be nil so the error check below is unreachable code.

mux_httpserver_test.go
```diff
@@ -48,3 +134,78 @@ func TestSchemeMatchers(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	// We create a ResponseRecorder (which satisfies http.ResponseWriter) to record the response.
+	rr := httptest.NewRecorder()
+	spa := spaHandler{staticPath: "./build", indexPath: "index.html"}
+	spa.FilepathAbsServeHTTP(rr, req)
+
+	status := rr.Code
+	if runtime.GOOS != "windows" && status != http.StatusOK {
```
elithrar, line 161: Don't create tests like this. Use build flags please.

elithrar (changes requested):
This isn't really mergeable.

* the example has unreachable code
* we add a number of dependencies to our tests, which don't test any code changes at all. At best, these should be in a self-contained `_example.go` file, separate from our tests. We aren't testing mux itself here.

README.md
amustaque97: OOPS! I missed it there. I think I forgot to remove the below if condition. Thank you for pointing it out.

amustaque97: Addressed the above review comment.

README.md
```diff
@@ -251,20 +251,11 @@ type spaHandler struct {
...
-	if err != nil {
-        // if we failed to get the absolute path respond with a 400 bad request
-        // and stop
-		http.Error(w, err.Error(), http.StatusBadRequest)
-		return
-	}
-
-    // prepend the path with the path to the static directory
+    // Join internally call path.Clean to prevent directory traversal
 	path = filepath.Join(h.staticPath, path)
```
ianling, line 255: `path` is never initialized

amustaque97: thanks for pointing it out. I have fixed the issue.

coreydaley (approved):

PR #697, opened 2022-09-18
 rename variable because collides with imported package name and remove unused parameters

Kamandlou:


mux.go
```diff
@@ -174,12 +174,12 @@ func (r *Router) Match(req *http.Request, match *RouteMatch) bool {
 // mux.Vars(request).
 func (r *Router) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	if !r.skipClean {
-		path := req.URL.Path
```
amustaque97, line 177: @Kamandlou, could you please share with me a code example where you're facing this collision? This makes me curious 🤔

Kamandlou: Variable name `path := req.URL.Path` has collide with imported package `import "path"`

amustaque97: As long as it is not a keyword, It is okay. Also, I noticed there were changes related to spaces/tabs. As long as `gofmt` doesn't complain about it, I don't see any issue with the indentation. I believe we can close this PR as this doesn't solve any existing problem or contribute to any enhancements. I hope you understand what I'm trying to say. Thank you

Issue #694, opened 2022-09-12
[bug]  Using route.Methods(...) more than once


edgy-sphere:
**Describe the bug**
Using `func (r *Route) Methods(methods ...string)` more than once for a single route results in a response with status code `405 Method Not Allowed` for this route.

**Background**
I tried to simplify the development of my web APIs, so I introduced some utility functions, with the relevant part essentially being:

<pre>
func RegisterRoute(
    mux *mux.Router,
    path string,
    methods []string,
    handler func(http.ResponseWriter, *http.Request),
  ) {
    mux.HandleFunc(path, handler).Methods("OPTIONS").Methods(methods...)
  }
</pre>

(`OPTIONS` is basically always required; `methods` as a slice because `PUT`, `PATCH` and even `POST` may point to the same handler)

**Versions**
go version go1.19 windows/amd64
> package version: run `git rev-parse HEAD` inside the repo -- what repo? I used go to get `mux@v1.8.0`

**Steps to Reproduce**
([GitHub repo](https://github.com/edgy-sphere/debug-mux))
<pre>
package main

import (
	"log"
	"net/http"

	"github.com/gorilla/mux"
)

func main() {
	mux := mux.NewRouter()
	mux.HandleFunc("/test", handler).Methods("PUT").Methods("PATCH")

	server := &http.Server{
		Addr: ":9710",
		Handler: mux,
	}

	err := server.ListenAndServe()
	if err != nil {
		log.Fatal(err)
	}
}

func handler(w http.ResponseWriter, r *http.Request) {
	w.WriteHeader(200)
}
</pre>

**Expected behavior**

Response with status code 200 (for the latter example).

**Solutions?**
Either:
- Disallow using `Methods(...)` more than once, e.g. via changing field `routeConf.matchers` from type `[]matcher` to type `map[string]matcher`, where key `string` (or any other practical type) may be `"METHODS"`.
- Force method matchers to be unique and to be merged if it already exists, e.g. as per above.
- Change `func (r *Route) Match(req *http.Request, match *RouteMatch)` to not return `ErrMethodMismatch` if _any_ method does not match, i.e. if there are multiple method matchers.

It has been not clear to me that using `Methods(...)` twice leads to this behaviour (hence this issue), so I would at least appreciate some kind of information at the function description in the source file -- although _I_ do know now.
Also, I neither fully understand your intended design nor Git things like Pull Requests, so I am sorry if my provided solutions may very well be rather lacking.

[...]

PR #699, opened 2022-10-09
Additional documentation for func Methods

henrikac:
Reading through #694 it sounds like that the behaviour of chaining multiple `func Methods(methods ...string)`is not clear. This PR is not a fix to this issue but it attempts to help make this behaviour more clear to other/new users.

**Summary of Changes**

1. A small note to `func Methods(methods ...string)` that explains the behavior of chaining multiple `Methods`.
2. Updated the methods section in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ r.PathPrefix("/products/")
 
 ```go
 r.Methods("GET", "POST")
+// This is equivalent to .Methods("GET")
+r.Methods("GET", "POST").Methods("GET")
 ```
 
 ...or URL schemes:
--- a/route.go
+++ b/route.go
@@ -322,6 +322,10 @@ func (m methodMatcher) Match(r *http.Request, match *RouteMatch) bool {
 // Methods adds a matcher for HTTP methods.
 // It accepts a sequence of one or more methods to be matched, e.g.:
 // "GET", "POST", "PUT".
+//
+// NOTE: Chaining multiple Methods will result in the last added
+// Methods will override the allowed methods specified by the
+// previously Methods calls.
 func (r *Route) Methods(methods ...string) *Route {
 	for k, v := range methods {
 		methods[k] = strings.ToUpper(v)
--- a/doc.go
+++ b/doc.go
@@ -81,6 +81,8 @@ There are several other matchers that can be added. To match path prefixes:
 ...or HTTP methods:
 
 	r.Methods("GET", "POST")
+	// This is equivalent to .Methods("GET")
+	r.Methods("GET", "POST").Methods("GET")
 
 ...or URL schemes:
 


route.go
```diff
@@ -322,6 +322,10 @@ func (m methodMatcher) Match(r *http.Request, match *RouteMatch) bool {
 // Methods adds a matcher for HTTP methods.
 // It accepts a sequence of one or more methods to be matched, e.g.:
 // "GET", "POST", "PUT".
+//
+// NOTE: Chaining multiple Methods will result in the last added
+// Methods will override the allowed methods specified by the
```
amustaque97, line 327: I don't think this is not the actual behaviour, have you tried this?  when I write something like the below:
```golang
r := mux.NewRouter()
r.HandleFunc("/health", healthHandle).Methods("POST").Methods("GET")
```

When I curl `/health` endpoint, it returns `HTTP ERROR 405`

README.md
```diff
@@ -107,6 +107,8 @@ r.PathPrefix("/products/")
 
 ```go
 r.Methods("GET", "POST")
+// This is equivalent to .Methods("GET")
```
amustaque97, line 110: This looks incorrect to me :(

Because when we write multiple `Methods` calls like: `Methods("GET", "POST").Methods("PATCH")`

Apparently, In one of the iterations [here](https://github.com/gorilla/mux/blob/master/route.go#L52) `matchErr` is set `ErrMethodMismatch` and it will return false. I hope you understand what I'm trying to say here.

amustaque97 (commented):
Hi @henrikac I left a few comments. Would you mind taking look at them? Looking forward to hearing from you.

route.go
henrikac: @amustaque97 this is weird but I am getting `405` for both the `POST` and `GET` request :thinking:

README.md
henrikac: Okay, `.Methods(methods ...string)` seems to act like a set and by chaining them makes them do some kind of intersection operation.

```go
.Methods("GET", "POST").Methods("PATCH") // will return 405 for all requests

.Methods("GET", "POST").Methods("GET", "PATCH") // will return 405 for all requests but GET requests
```

route.go
itsdeekay: @henrikac  this is expected behaviour. Consider it as series of filter. When you filter the requests based on POST and then try to do it with GET, the result will be 0 and hence 405 error

henrikac: > @henrikac this is expected behaviour. Consider it as series of filter. When you filter the requests based on POST and then try to do it with GET, the result will be 0 and hence 405 error

Yeah, I figured. In my comment below for the README I mention that I think it looks like it behaves like it performs an intersection operation :)

itsdeekay: @henrikac so is this PR required? Could you modify or decline PR based on changes/feedback

PR #731, opened 2023-08-24
Add RegexpCompileFunc to override regexp.Compile

titpetric:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Route creation is slow as it always invokes regexp.Compile, increasing memory and cpu pressure in cases where route creation is common (e.g. an api gateway reloading changes). This PR adds a exported `RegexpCompileFunc` variable, that's possible to override with other implementations, adding caches. Added a benchmark.

```
# go test -bench=NewRouter -benchmem -memprofile memprofile.out -cpuprofile profile.out -count 10
goos: linux
goarch: amd64
pkg: github.com/gorilla/mux
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkNewRouter-8             	   36786	     34450 ns/op	   26327 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   44839	     31840 ns/op	   26327 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   48939	     23419 ns/op	   26327 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   52255	     25949 ns/op	   26327 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   44684	     23150 ns/op	   26329 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   41522	     35047 ns/op	   26328 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   48354	     35461 ns/op	   26329 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   33850	     38467 ns/op	   26330 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   36920	     33917 ns/op	   26330 B/op	     374 allocs/op
BenchmarkNewRouter-8             	   40767	     29980 ns/op	   26329 B/op	     374 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  303121	      3738 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  311847	      4184 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  300144	      4212 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  329636	      4528 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  277980	      5136 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  363416	      4578 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  328996	      4231 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  327628	      3699 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  300370	      3955 ns/op	    2161 B/op	      73 allocs/op
BenchmarkNewRouterRegexpFunc-8   	  297345	      4115 ns/op	    2161 B/op	      73 allocs/op
PASS
ok  	github.com/gorilla/mux	35.091s
```

Benchmark: NewRouter

Average runtime per iteration: ~34450 ns/op
Average memory usage per iteration: ~26327 B/op
Average allocations per iteration: ~374 allocs/op

Benchmark: NewRouterRegexpFunc

Average runtime per iteration: ~4184 ns/op (approx. 87.8% improvement)
Average memory usage per iteration: ~2161 B/op (approx. 91.8% reduction)
Average allocations per iteration: ~73 allocs/op (approx. 80.5% reduction)

As per go doc, [using *regexp.Regexp concurrently is safe](https://pkg.go.dev/regexp#Regexp).

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [ ] `make verify` is passing
- [x] `make test` is passing

### DIFF
--- a/route_test.go
+++ b/route_test.go
@@ -57,7 +57,7 @@ func BenchmarkNewRouterRegexpFunc(b *testing.B) {
 	}
 }
 
-func testNewRouter(tb testing.TB, handler http.Handler) {
+func testNewRouter(_ testing.TB, handler http.Handler) {
 	r := NewRouter()
 	// A route with a route variable:
 	r.Handle("/metrics/{type}", handler)
--- a/route_test.go
+++ b/route_test.go
@@ -1,0 +1,66 @@
+package mux
+
+import (
+	"net/http"
+	"regexp"
+	"sync"
+	"testing"
+)
+
+var testNewRouterMu sync.Mutex
+var testHandler = http.NotFoundHandler()
+
+func BenchmarkNewRouter(b *testing.B) {
+	testNewRouterMu.Lock()
+	defer testNewRouterMu.Unlock()
+
+	// Set the RegexpCompileFunc to the default regexp.Compile.
+	RegexpCompileFunc = regexp.Compile
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		testNewRouter(b, testHandler)
+	}
+}
+
+func BenchmarkNewRouterRegexpFunc(b *testing.B) {
+	testNewRouterMu.Lock()
+	defer testNewRouterMu.Unlock()
+
+	// We preallocate the size to 8.
+	cache := make(map[string]*regexp.Regexp, 8)
+
+	// Override the RegexpCompileFunc to reuse compiled expressions
+	// from the `cache` map. Real world caches should have eviction
+	// policies or some sort of approach to limit memory use.
+	RegexpCompileFunc = func(expr string) (*regexp.Regexp, error) {
+		if regex, ok := cache[expr]; ok {
+			return regex, nil
+		}
+
+		regex, err := regexp.Compile(expr)
+		if err != nil {
+			return nil, err
+		}
+
+		cache[expr] = regex
+		return regex, nil
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		testNewRouter(b, testHandler)
+	}
+}
+
+func testNewRouter(tb testing.TB, handler http.Handler) {
+	r := NewRouter()
+	// A route with a route variable:
+	r.Handle("/metrics/{type}", handler)
+	r.Queries("orgID", "{orgID:[0-9]*?}")
+	r.Host("{subdomain}.domain.com")
+}
--- a/mux.go
+++ b/mux.go
@@ -19,6 +19,10 @@ var (
 	ErrMethodMismatch = errors.New("method is not allowed")
 	// ErrNotFound is returned when no route match is found.
 	ErrNotFound = errors.New("no matching route was found")
+	// RegexpCompileFunc aliases regexp.Compile and enables overriding it.
+	// Do not run this function from `init()` in importable packages.
+	// Changing this value is not safe for concurrent use.
+	RegexpCompileFunc = regexp.Compile
 )
 
 // NewRouter returns a new router instance.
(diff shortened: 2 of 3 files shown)


coreydaley:
@titpetric Thank you for the pull request! We will need a little bit of time to digest this and review it, but it is definitely towards the top of our queue.

titpetric:
@coreydaley Let me know if there's absolutely anything I can do to improve it. I'm on a weird precipice where CSEs are routing for using the internal fork and I'd like to avoid that.

I've considered adding a .SetRegexpCompile(func(...)...) as an API addition, but wen't with the 'less is more' approach. It would be possible to update this in runtime with some mutex protections. That sounds fine until you consider how many extra bugs can rise up when conccurency is at play.

Explicit control of the caching behaviour is a single shot - if someone wants to control that during the router lifetime, just implementing the compile func signature enables it. It's expected there would be some sort of smarter eviction implementations behind it. Overthinking it just leads to more unnecessary problems, keeping things simple is hard sometimes.

AlexVulaj:
@coreydaley This looks good to me, though I'll defer to you for a merge since I know you were looking at it first.

While I'm in favor of giving consumers the freedom to optimize their route creation, I believe such an action comes with the implicit disclaimer that if someone is going to play with around with it, they should know what they're doing and are on their own if it breaks something.

route_test.go
AlexVulaj, line 1: I think this file is good to have as documentation and proof of the benefit of this change, though it feels odd to have as a `_test.go` file that will run on every build since it's not actually asserting anything and can't fail. I'm wondering if there's either

a.) a better place we can keep it
or
b.) any assertions we can add to it that would be worth testing on every build

Interested in hearing your thoughts.

route_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		testNewRouter(b, testHandler)
+	}
+}
+
+func testNewRouter(tb testing.TB, handler http.Handler) {
```
AlexVulaj, line 60:
```suggestion
func testNewRouter(handler http.Handler) {
```

I don't see this param used anywhere, so I'm in favor of cleaning it up.

AlexVulaj:
Another note - all of the changes from this PR seem to also be present in https://github.com/gorilla/mux/pull/732

route_test.go
AlexVulaj: Actually scratch this - I see that they all start with "Benchmark" so we should be good here. My apologies 😃

route_test.go
titpetric: Can we come up with an assertion here, assert on some value of Router, Route, Walk...?

Alternatively, I'd use `_ testing.TB` to make it clear that the parameter is (yet) unused, leaving it in place for assertions in the future.

titpetric:
> Another note - all of the changes from this PR seem to also be present in #732

Correct, regexp is it's own optimization scope (this PR). The second PR needs a rebase when this is merged, that addresses actual memory pressure, while this one just gives us an ability to use a cached/compiled *regexp.Regexp.

titpetric:
- rebased
- marked testing.TB internal func param as `_` (leave in place)

route_test.go
AlexVulaj: Seeing as these are all benchmark tests, I'm content with not asserting on anything at this time.

AlexVulaj (approved):

PR #732, opened 2023-08-24
Optimize/new route regexp allocations

titpetric:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

This optimizes NewRouter creation. The change introduces strings.Builder for more efficient strings concatenation, and replaces SplitN with Index, avoiding the allocations for the slice.

| Test Group                  | Condition | Average Time (ns/op) | Average Memory (B/op) | Average Allocations |
|-----------------------------|-----------|----------------------|-----------------------|---------------------|
| **BenchmarkNewRouter**      | Before    | 37041                | 26553                  | 376                 |
|                             | After     | 32046                | 25800                  | 347                 |
| **BenchmarkNewRouterRegexpFunc** | Before | 4713                 | 2385                   | 75                  |
|                             | After     | 2792                 | 1640                   | 46                  |

BenchmarkNewRouterRegexpFunc:

Time Improvement: Significant improvement in execution time from 4,713 ns/op to 2,792 ns/op.
Memory Usage: Memory usage per operation dropped from 2,385 bytes to 1,640 bytes.
Allocations: Allocations per operation decreased substantially from 75 to 46.

## Added/updated tests?

- [x] Yes

## Run verifications and test

- [ ] `make verify` is passing
- [x] `make test` is passing

### DIFF
--- a/regexp.go
+++ b/regexp.go
@@ -65,35 +65,48 @@ func newRouteRegexp(tpl string, typ regexpType, options routeRegexpOptions) (*ro
 	}
 	varsN := make([]string, len(idxs)/2)
 	varsR := make([]*regexp.Regexp, len(idxs)/2)
-	pattern := bytes.NewBufferString("")
+
+	var pattern, reverse strings.Builder
 	pattern.WriteByte('^')
-	reverse := bytes.NewBufferString("")
-	var end int
+
+	var end, colonIdx, groupIdx int
 	var err error
+	var patt, param, name string
 	for i := 0; i < len(idxs); i += 2 {
 		// Set all values we are interested in.
+		groupIdx = i/2
+
 		raw := tpl[end:idxs[i]]
 		end = idxs[i+1]
-		parts := strings.SplitN(tpl[idxs[i]+1:end-1], ":", 2)
-		name := parts[0]
-		patt := defaultPattern
-		if len(parts) == 2 {
-			patt = parts[1]
+
+		param = tpl[idxs[i]+1 : end-1]
+		colonIdx = strings.Index(param, ":")
+		if colonIdx == -1 {
+			name = param
+			patt = defaultPattern
+		} else {
+			name = param[0:colonIdx]
+			patt = param[colonIdx+1:]
 		}
+		if patt == "" {
+			patt = defaultPattern
+		}
+
 		// Name or pattern can't be empty.
 		if name == "" || patt == "" {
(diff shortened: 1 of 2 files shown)


titpetric:
Rebased off #731

titpetric:
@AlexVulaj thanks for #731 merge; rebased this one for review. Benchmark still the same as in description.

titpetric:
@coreydaley is there anything i can/should do to get this reviewed?

regexp.go
```diff
@@ -65,67 +65,88 @@ func newRouteRegexp(tpl string, typ regexpType, options routeRegexpOptions) (*ro
...
+		}
+		if patt == "" {
+			patt = defaultPattern
 		}
+
 		// Name or pattern can't be empty.
 		if name == "" || patt == "" {
-			return nil, fmt.Errorf("mux: missing name or pattern in %q",
-				tpl[idxs[i]:end])
+			return nil, fmt.Errorf("mux: missing name or pattern in %q", param)
```
apoorvajagtap, line 97:
```suggestion
			return nil, fmt.Errorf("mux: missing name or pattern in %q", tpl[idxs[i]:end])
```
Having `param` as the variable might not help much, as when there's no pattern or name inside braces {}, param will just return an empty string (""), whereas. Above will at least help with the braces for reference.

apoorvajagtap (changes requested):

apoorvajagtap:
Hi @titpetric, thanks for working on this! 
Also, thanks for being patient while we were adapting to the changes :). I have looked through & added some inputs, please feel free to share your thoughts.

regexp.go
titpetric: Gotcha, so {} would be printed as well. Sounds legit, I'll give it a once over and add a test, some inline comments, it's a bit hard to read after some time since writing it

apoorvajagtap:
@titpetric LGTM.
Would you like to squash the commits please? (mainly those that are very related and just have slight differences)

apoorvajagtap (approved):

apoorvajagtap:
The lint & security scan errors are not related to the current changes, so I'll go ahead with squash & merge.

mdnix:
Will there be a new release soon that includes this change? My team and I came across this because we specifically need this type of optimization.

PR #746, opened 2023-12-19
fix(#739): Remove else statement on route mismatch

### LINKED ISSUE #739: [BUG] MethodNotAllowedHandler does not work for subrouter with different routes

- [X] I have searched the existing issues

### Current Behavior

> Go version: 1.20
    Mux version: 1.8.1

Starting with mux `v1.8.1` we encountered the following problem, which was not present in previous versions of the library.
If there are several routes in one subrouter, for example `/api` and `/api/{id}` MethodNotAllowedHandler does not work and NotFoundHandler is used by default.

An example test to reproduce the bug is shown below. The bug is not reproduced if:
- no subrouter is used, just main mux.Router
- subrouter contains only routes of the same type (only `/api` for example)
- place HandleFunc for `/api/{id}` before HandleFunc for `/api` in the code

After examining the modified code, it was found out that the problem is related to the addition of this else condition:
https://github.com/gorilla/mux/blob/v1.8.1/route.go#L70-L78
Since now the result of the Route.Match function depends on which Route was processed last (the value of match.MatchErr by reference is updated).

### Expected Behavior

Result 405 Method Not Allowed

### Steps To Reproduce

1. Run test:
```
func TestMiddlewareMethodMismatch(t *testing.T) {
	router := mux.NewRouter()
	router.MethodNotAllowedHandler = http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
		rw.WriteHeader(http.StatusMethodNotAllowed)
		_, _ = rw.Write([]byte("Method not allowed"))
	})

	subrouter := router.PathPrefix("/v1").Subrouter()
	subrouter.HandleFunc("/api", func(w http.ResponseWriter, e *http.Request) {
		_, _ = w.Write([]byte("Logic"))
	}).Methods("GET")
	subrouter.HandleFunc("/api/{id}", func(w http.ResponseWriter, e *http.Request) {
		_, _ = w.Write([]byte("Logic"))
	}).Methods("GET")

	rw := httptest.NewRecorder()
	req, _ := http.NewRequest("POST", "/v1/api", nil)
	router.ServeHTTP(rw, req)

	if rw.Code == http.StatusNotFound {
		t.Fatal("Not found instead of Method Not Allowed")
	}
}
```

### Anything else?

_No response_

marcelom97:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description

When a path with specific query params does not match we should return a ErrMethodMismatch instead of ErrNotFound

- Related Issue #739 
- Closes #739 

## Added/updated tests?

- [X] Yes
      have not been included_

## Run verifications and test

- [X] `make verify` is passing
- [X] `make test` is passing

### DIFF
--- a/mux_test.go
+++ b/mux_test.go
@@ -2095,8 +2095,8 @@ func TestMultipleDefinitionOfSamePathWithDifferentMethods(t *testing.T) {
 		if matched {
 			t.Error("Should not have matched route for methods")
 		}
-		if match.MatchErr != ErrNotFound {
-			t.Error("Should have ErrNotFound error. Found:", match.MatchErr)
+		if match.MatchErr != ErrMethodMismatch {
+			t.Error("Should have ErrMethodMismatch error. Found:", match.MatchErr)
 		}
 	})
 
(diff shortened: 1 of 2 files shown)


jaitaiwan:
I've reviewed and I'm happy with the code and tests to match. What needs to be considered now is if this is technically a breaking api change or not as folks utilising the library may expect the former error. @AlexVulaj would love your input on this.

jaitaiwan:
Actually I got confused - this PR is now obsolete. Thank you heaps for the contribution @marcelom97. If you feel that this was in error, please open up another issue to discuss.

PR #761, opened 2024-05-31
Feature router in request context

cornejong:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
Named routes are great, and being able to retrieve the URL for a route based on its name is a feature I use a lot. However, when handlers reside in a different package from the router instance, this can become cumbersome. Therefore, it would be great if we could fetch the current router just like we can fetch the current route. This PR adds that functionality.

```go
router := mux.CurrentRouter(r)
```
By default, the router is added to the request context but can be omitted using the `omitRouterFromContext` option or by calling `router.OmitRouterFromContext(true)`.

- Related Issue #
- Closes #

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [x] `make verify` is passing
- [x] `make test` is passing

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -286,6 +286,15 @@ func (r *Router) OmitRouteFromContext(value bool) *Router {
 	return r
 }
 
+// OmitRouterFromContext defines the behavior of omitting the Router from the
+// http.Request context.
+//
+// RouterFromRequest will yield nil with this option.
+func (r *Router) OmitRouterFromContext(value bool) *Router {
+	r.omitRouterFromContext = value
+	return r
+}
+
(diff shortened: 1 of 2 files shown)


AlexVulaj:
I really like this change - I foresee some good use cases in middleware or anything that handles multiple routes. Thanks for the contribution!

AlexVulaj (approved):

cornejong:
Thanks for the merge! Always happy to contribute.

PR #764, opened 2024-06-20
Feature/method case insensitive

### LINKED ISSUE #762: [FEATURE] Ignore case sensitivity when matching defined methods
### Is there an existing feature request for this?

- [X] I have searched the existing feature requests

### Is your feature request related to a problem? Please describe.

If a client is (mis)configured to make requests with lowercase method names, it is impossible to use gorilla/mux to match.

### Describe the solution that you would like.

Case insensitive matches on methods.

### Describe alternatives you have considered.

One way is to stop `strings.ToUpper`-ing defined methods, but that would break compatibility. Documenting the current limitation if the behavior is not changed would also go far.

### Anything else?

A test case that demonstrates the condition:
```go
func TestLowercaseMethods(t *testing.T) {
	r := mux.NewRouter()

	r.HandleFunc("/hc", func(w http.ResponseWriter, r *http.Request) {
		w.WriteHeader(http.StatusOK)
	}).Methods("get", "GET")

	methods := []string{"get", "GET"}

	for i := range methods {
		w := httptest.NewRecorder()
		r.ServeHTTP(w, httptest.NewRequest(methods[i], "/hc", nil))
		t.Logf("Method %q: got %d", methods[i], w.Result().StatusCode)
	}
}

// Method "get": got 405
// Method "GET": got 200
```

Also, thanks!

cornejong:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Go Version Update
- [ ] Dependency Update

## Description
This PR adds the ability to define case insensitive method matching for routes. as suggested in issue #762.
To achieve this this pr adds the following:
```GO
// methodMatcher matches the request against HTTP methods without case sensitivity.
// Both the supplied methods as well as the request method will be transformed to uppercase.
type methodCaseInsensitiveMatcher []string

// methodCaseExactMatcher matches the request against HTTP methods exactly.
// No transformation of supplied methods or the request method is applied.
type methodCaseExactMatcher []string

// Explicitly apply the default (and current)  method matcher, overrides the router default
func (r *Route) MethodsDefault(methods ...string) *Route
// Explicitly apply the case insensitive method matcher, overrides the router default
func (r *Route) MethodsCaseInsensitive(methods ...string) *Route
// Explicitly apply the exact method matcher, overrides the router default
func (r *Route) MethodsCaseExact(methods ...string) *Route

// Set routeConf value for methodMatcher, for this route, to methodDefaultMatcher
func (r *Route) MatchMethodDefault() *Route
// Set routeConf value for methodMatcher, for this route, to methodCaseInsensitiveMatcher
func (r *Route) MatchMethodCaseInsensitive() *Route
// Set routeConf value for methodMatcher, for this route, to methodCaseExactMatcher
func (r *Route) MatchMethodCaseExact() *Route

// Set default method matcher for the router to methodDefaultMatcher
func (r *Router) MatchMethodDefault() *Router
// Set default method matcher for the router to methodCaseInsensitiveMatcher
func (r *Router) MatchMethodCaseInsensitive() *Router
// Set default method matcher for the router to methodCaseExactMatcher
func (r *Router) MatchMethodCaseExact() *Router


type routeConf struct {
    // ...
    // Holds the default method matcher
    methodMatcher matcher
}

// as well a the helper
func sliceToUpper(slice []string) []string 
```
### Usage
Setting the default method matcher on the router:
```go
r := mux.NewRouter()
r.MatchMethodDefault()          // For the default matcher
r.MatchMethodCaseInsensitive()  // For case insensitive matching
r.MatchMethodCaseExact()        // For exact matching

r.HandleFunc("/", func).Methods("GET") //  will use the default matcher set on the router
```
Applying a specific matcher to a single route and overriding the default method matcher of the router:
```go
r.HandleFunc("/", func).MatchMethodDefault().Methods("GET")
r.HandleFunc("/", func).MatchMethodCaseInsensitive().Methods("GET")
r.HandleFunc("/", func).MatchMethodCaseExact().Methods("GET")
// Not the prefferd method, order of opperations is critcal here. 
// Since the Methods(...) func adds the matcher. 
// Invoking the MatchMethodCaseInsensitive() method after it will have no effect.

// OR, the more preferred method:

r.HandleFunc("/", func).MethodsDefault("GET")
// Explicitly match the methods for this route with the default matcher
r.HandleFunc("/", func).MethodsCaseInsensitive("GET")
// Explicitly match the methods for this route without case sensitivity
r.HandleFunc("/", func).MethodsCaseExact("GET")
// Explicitly match the methods for this route exactly

```

While using the standard ```Methods(...)``` method the default matcher for the router will be used. 

- Related Issue #762 
- Closes #762 

## Added/updated tests?

- [x] Yes
      have not been included_

## Run verifications and test

- [x] `make verify` is passing
- [x] `make test` is passing

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -29,7 +29,12 @@ var (
 
 // NewRouter returns a new router instance.
 func NewRouter() *Router {
-	return &Router{namedRoutes: make(map[string]*Route)}
+	return &Router{
+		namedRoutes: make(map[string]*Route),
+		routeConf: routeConf{
+			methodMatcher: methodDefaultMatcher{},
+		},
+	}
 }
 
 // Router registers routes to be matched and dispatches a handler.
@@ -104,9 +109,8 @@ type routeConf struct {
 
 	buildVarsFunc BuildVarsFunc
 
-	// If true, methods will be matched case insensitive.
-	// The methodCaseInsensitiveMatcher will be used instead of methodMatcher
-	matchMethodCaseInsensitive bool
+	// Holds the default method matcher
+	methodMatcher matcher
 }
 
 // returns an effective deep copy of `routeConf`
@@ -282,12 +286,6 @@ func (r *Router) OmitRouteFromContext(value bool) *Router {
 	return r
 }
 
-// MatchMethodCaseInsensitive defines the behaviour of ignoring casing for request methods.
-func (r *Router) MatchMethodCaseInsensitive(value bool) *Router {
-	r.matchMethodCaseInsensitive = value
-	return r
-}
-
 // UseEncodedPath tells the router to match the encoded original path
 // to the routes.
 // For eg. "/path/foo%2Fbar/to" will match the path "/path/{var}/to".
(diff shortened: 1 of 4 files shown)


ghost:
[HTTP method names are case sensitive](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.1). "GET" and "get" are different methods.

This PR adds a large amount of API surface area for something that should rarely be used.

cornejong:
> [HTTP method names are case sensitive](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.1). "GET" and "get" are different methods.
> 
> 
> 
> This PR adds a large amount of API surface area for something that should rarely be used.

@hulkingshtick, actually I totally agree with you and no longer think this is a good approach.

jaitaiwan:
I too agree, closing

PR #810, opened 2026-08-31
Do not panic when routeRegexp.regexp is nil

### LINKED ISSUE #749: [BUG] runtime error in (*routeRegexp).Match

- [X] I have searched the existing issues

### Current Behavior

Code panics

### Expected Behavior

Not to panic but return an error

### Steps To Reproduce

It's a sporadic issue we spotted through automation (Sentry) so hard to tell how it could be reproduced but somehow the regexp pointer is `nil` when the code attempts to match.

### Anything else?

Stacktrace

```
runtime.errorString: runtime error: invalid memory address or nil pointer dereference
  File "regexp/exec.go", line 527, in (*Regexp).doExecute
  File "regexp/exec.go", line 514, in (*Regexp).doMatch
  File "regexp/regexp.go", line 533, in (*Regexp).MatchString
  File "github.com/gorilla/mux@v1.8.0/regexp.go", line 193, in (*routeRegexp).Match
  File "github.com/gorilla/mux@v1.8.0/route.go", line 50, in (*Route).Match
  File "github.com/gorilla/mux@v1.8.0/mux.go", line 138, in (*Router).Match
  File "github.com/gorilla/mux@v1.8.0/route.go", line 50, in (*Route).Match
  File "github.com/gorilla/mux@v1.8.0/mux.go", line 138, in (*Router).Match
  File "github.com/gorilla/mux@v1.8.0/route.go", line 50, in (*Route).Match
  File "github.com/gorilla/mux@v1.8.0/mux.go", line 138, in (*Router).Match
  File "github.com/gorilla/mux@v1.8.0/mux.go", line 196, in (*Router).ServeHTTP
  File "net/http/server.go", line 2109, in HandlerFunc.ServeHTTP
  File "net/http/server.go", line 2947, in serverHandler.ServeHTTP
  File "net/http/server.go", line 1991, in (*conn).serve
```

AshSgDe29071999:
Fixes #749

`routeRegexp.Match` called `regexp.MatchString` without checking the compiled pointer. Production Sentry traces showed a nil `*Regexp` panicking in `Regexp.doExecute` during `Router.ServeHTTP`.

Treat a nil `routeRegexp` or a nil compiled regexp as no match instead of panicking.


mux_test.go
```diff
@@ -3239,3 +3239,15 @@ func newRequestHost(method, url, host string) *http.Request {
...
 	return req
 }
+
+func TestRouteRegexpMatchNilRegexp(t *testing.T) {
+	req := newRequest("GET", "http://example.com/foo")
+	var r *routeRegexp
+	if r.Match(req, &RouteMatch{}) {
+		t.Fatal("nil routeRegexp should not match")
+	}
+	r = &routeRegexp{regexpType: regexpTypePath}
```
thaJeztah, line 3249: This is manually constructing the `routeRegexp`, which is a type that's non-exported and has a constructor; https://github.com/gorilla/mux/blob/db9d1d0073d27a0a2d9a8c1bc52aa0af4374d265/regexp.go#L31-L41


I'm not sure if fixing this by adding guards is the right thing to do; the related ticket indicates that either something is broken elsewhere, or that the code is used incorrectly, leading to the panic (which may actually be the correct thing to do for such cases);

- https://github.com/gorilla/mux/issues/749

